PR #7, opened 2017-05-11
Deploy netbox via stable tarball releases

blemagicleap:
Hi Musee, 
Please review my latest updated of the deploy netbox via stable tarball. I will work on the next issue regard to "template configuration".

Ben


tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - epel-release
+    - ca-certificates
+    - sudo
+    - python-virtualenv
+    - tar 
+
+- name: Download and untar netbox tarball
```
lae: This task is duplicated amongst all of the deploy tasks file with no change - they should probably be in a single shared tasks file, otherwise plays using this role will be polluted with duplicate tasks (which makes it hard to use the `--start-at-task` parameter from `ansible-playbook`) because Ansible will load all included tasks file regardless of if a conditional is false or true (if false, it'll just print that the task is skipped).

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+  unarchive:
+    src: "{{ netbox_stable_uri }}"
+    dest: "{{ netbox_releases_path }}"
+    copy: no
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+
+- name: Create a link to active release
```
lae: This is also a task that should be deduplicated.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: Install pip3 
+  shell: /usr/bin/easy_install-3.4 pip
+
+- name: Install virtualenv module for pip3
+  pip: 
+    name: virtualenv
+    state: present
+
+- name: Install Requirement Python Packages
```
lae: This is also a task that should be deduplicated. Also, I'd rename this task to `Install NetBox python dependencies inside active virtualenv`

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+  pip: 
+    requirements: "{{ netbox_current_path }}/requirements.txt"
+    virtualenv: "{{ netbox_current_path }}/venv"
+    virtualenv_python: python3
```
lae: You're hard coding Python 3 here. This task should make use of the `netbox_python` variable and work for deploying NetBox with both Python 2 and 3. So maybe something like:

    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
    virtualenv_python: "/usr/bin/python{{ netbox_python }}"

tasks/check_mandatory_vars.yml
```diff
@@ -11,16 +11,3 @@
...
-    msg: "Please define either database_socket (local) or database_host/database_password (TCP)."
-  when:
-    - netbox_database_socket is not defined
-    - netbox_database_host is not defined or (netbox_database_host is defined and netbox_database_password is not defined)
-
-- name: Ensure only one DB connection method is defined
-  fail:
-    msg: "Please define only one database connection method."
-  when:
-    - netbox_database_socket is defined and netbox_database_host is defined
```
lae: Please run:

    git checkout master tasks/check_mandatory_vars.yml

in your repository directory and commit it to restore this file correctly.

tasks/debian_deploy.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Install python-virtualenv
+  apt:
+    name: python-virtualenv
+    state: latest
```
lae: Should be a dependency listed in `vars/debian.yml`

tasks/debian_deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: Create a link to active release
+  file: 
+    src: "{{ netbox_releases_path }}/netbox-{{ netbox_stable_release }}"
+    dest: "{{ netbox_current_path }}"
+    state: link
+
+- name: Install pyparsing and appdirs modules via pip
+  pip:
+    virtualenv: "{{ netbox_current_path }}/venv"
+    virtualenv_python: python3
```
lae: Same issue as described above with not supporting Python 2

tasks/debian_deploy_via_git.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Install git package
+  apt: 
+    name: git
+    state: latest
+
+- name: Clone Netbox from Github
+  git: 
+    repo: https://github.com/digitalocean/netbox.git
```
lae: Repository should not be hard coded, this is already defined in `netbox_git_uri` to allow users to deploy their own NetBox forks.

tasks/main.yml
```diff
@@ -48,25 +48,24 @@
...
-- name: Ensure Postgres database exists (via TCP)
-  postgresql_db:
-    name: "{{ netbox_database }}"
-    login_host: "{{ netbox_database_host }}"
-    login_port: "{{ netbox_database_port }}"
-    login_user: "{{ netbox_database_user }}"
-    login_password: "{{ netbox_database_password }}"
-  when:
-    - netbox_database_socket is not defined
-    - netbox_database_host is not defined
```
lae: All of these tasks should be restored, probably best if you do it manually (`git checkout` won't work here since you made changes to this file).

tasks/main.yml
```diff
@@ -48,25 +48,24 @@
...
-    postgresql_db:
-      name: "{{ netbox_database }}"
-      login_user: "{{ netbox_database_user }}"
-      login_unix_socket: "{{ netbox_database_socket }}"
-    become: True
-    become_user: "{{ netbox_database_user }}"
-  when:
-    - netbox_database_socket is defined
-    - netbox_database_host is not defined
+- name: Create Directories
```
lae: I would rename this to `Create directories for NetBox deployment`

tasks/main.yml
```diff
@@ -48,25 +48,24 @@
...
+  file: 
+    path: "{{ item }}" 
+    state: directory
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+  with_items:
+    - "{{ netbox_releases_path }}"
+    - "{{ netbox_shared_path }}"
```
lae: This should also have `netbox_git_path`

tasks/ubuntu_deploy.yml
```diff
@@ -0,0 +1,26 @@
...
+    group: "{{ netbox_group }}"
+
+- name: Create a link to active release
+  file: 
+    src: "{{ netbox_releases_path }}/netbox-{{ netbox_stable_release }}"
+    dest: "{{ netbox_current_path }}"
+    state: link
+
+- name: Install Python virtualenv
+  apt:
```
lae: Should go in `vars/ubuntu.yml`.

vars/debian.yml
```diff
@@ -14,3 +14,5 @@ netbox_python3_packages:
   - python3
   - python3-dev
   - python3-pip
+  - libjpeg62-turbo-dev
```
lae: This only installs `libjpeg62-turbo-dev` when `netbox_python` is set to 3. This is not a Python package and should go under `netbox_packages`.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Install OS Requirement Packages
+  yum:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - epel-release
```
lae: Actually never mind this - EPEL is already installed in https://github.com/lae/ansible-role-netbox/blob/master/tasks/main.yml#L13 so this is a duplicate install task

tasks/main.yml
blemagicleap: I don't see these lines in the origin master tasks/mail.yml

vars/debian.yml
blemagicleap: Done

tasks/main.yml
lae: Try running `git fetch --all`? They are in master, unless you removed them from your master branch...

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Install OS Requirement Packages
```
blemagicleap: Removed ca-certificates, tar, and sudo packages.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+  yum:
+    name: "{{ item }}"
+    state: latest
+  with_items:
+    - epel-release
+    - ca-certificates
+    - sudo
+    - python-virtualenv
```
blemagicleap: Fixed. Moved python-virtual package from tasks/centos_deploy.yml to vars/redhat.yml

tasks/centos_deploy.yml
blemagicleap: I thought that we want to deploy the Netbox to OS and LXC containers as well.

lae: LXC containers can be supported, the tests pass on LXC containers without you installing these you know. Just, you're using a minimal, stripped down LXC template for /your/ tests - there's no reason for this role to, in a sense, install the base operating system. That's what you're kind of doing when you're installing these packages. It's expected that these are installed when using Ansible and connecting to external websites over HTTPS.

defaults/main.yml
```diff
@@ -4,7 +4,7 @@ netbox_stable: false
 netbox_stable_release: 1.9.6
 netbox_stable_uri: "https://github.com/digitalocean/netbox/archive/v{{ netbox_stable_release }}.tar.gz"
 
-netbox_git: false
+netbox_git: true
```
lae: Scenario should not be selected by default. Can you change this back to false?

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Install pip3 
+  shell: /usr/bin/easy_install-3.4 pip
+
+- name: Install virtualenv module for pip3
+  pip: 
+    name: virtualenv
+    state: present
+
+- include: install_py_deps_inside_venv.yml
```
lae: This is still duplicating a task by including it - it makes more sense to just include this in `main.yml` (include the task verbatim I mean, not via an include). Includes should really only be used to group similar tasks (e.g. for configuring a specific feature) - not for a single task.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Install pip3 
+  shell: /usr/bin/easy_install-3.4 pip
+
+- name: Install virtualenv module for pip3
+  pip: 
+    name: virtualenv
+    state: present
```
lae: This still looks like it's duplicating installation of the `python-virtualenv` rpm package specified in `vars/redhat.yml` but as a python package. Is this actually needed, and if so why?

tasks/main.yml
lae: Still missing this.

tasks/install_py_deps_inside_venv.yml
```diff
@@ -0,0 +1,5 @@
+- name: Install Netbox python dependencies inside active virtualenv
+  pip:
+    requirements: "{{ netbox_current_path }}/requirements.txt"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+    virtualenv_python: "/usr/bin/python{{ netbox_python }}"
```
lae: This task looks good, but like mentioned above it should probably go in `main.yml` and not as a separate tasks file.

tasks/ubuntu_deploy.yml
```diff
@@ -0,0 +1,2 @@
+---
+- include: install_py_deps_inside_venv.yml
```
lae: This is a good example of the silly things one can do when violating the DRY (Don't Repeat Yourself) principle. It doesn't quite make sense to have a task file that just includes another task file and does nothing else.

lae:
Also, it sounds like it would benefit if a `.gitignore` file was included to ignore the `.swp` files from vim. I'm not sure how you added the `FETCH_HEAD` file but that should be removed.

lae:
I think given the current changeset, it makes sense to not have a "deploy.yml" task file per distribution, but leave everything in `main.yml` and test for distribution for the minor differences on some distributions.

defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # defaults file for lae.netbox
-netbox_stable: false
+netbox_stable: true
```
lae: The default for this needs to be false - as stated before we should not have a scenario selected by default.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Install pip3 
+  shell: /usr/bin/easy_install-3.4 pip
+  changed_when: false
+  failed_when: false
+  always_run: true
```
lae: This is not what idempotence means - all you're doing here is telling Ansible to run easy_install every time this role is run, which actually does make changes on the system (hence, breaking the idempotency trait), and tricking Ansible with "oh I'm running this command and it doesn't change anything."

The better option is to use the `creates` argument for the `shell` module to indicate that when a file (that should be created by this shell task) exists, then it doesn't need to run the shell task again. I assume that could probably be `/usr/bin/pip` or whatever?

The other thing is that this is Python 3 specific - you need to support both Python 2 and 3, so you should modify this task so that it'll install pip using either Python 2 or 3, depending on what `netbox_python` is set to.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,11 @@
...
+
+- name: Install virtualenv module via pip
+  pip:
+    name: virtualenv
+    state: present
```
lae: This should probably select a specific Python binary depending on which Python version is selected (`netbox_python`).

templates/configuration.py.j2
```diff
@@ -0,0 +1,85 @@
...
+BANNER_TOP = ''
+BANNER_BOTTOM = ''
+
+# When determining the primary IP address for a device, IPv6 is preferred over IPv4 by default. Set this to True to
+# prefer IPv4 instead.
+PREFER_IPV4 = '{{ "True" if netbox_prefer_ipv4 else "False" }}'
+
+# Enforcement of unique IP space can be toggled on a per-VRF basis. To enforce unique IP space within the global table
+# (all prefixes and IP addresses not assigned to a VRF), set ENFORCE_GLOBAL_UNIQUE to True.
+ENFORCE_GLOBAL_UNIQUE = False
```
lae: You should leave this file out for this PR - it'll get confusing to have this merged here and then later have a separate commit that adds the variables that are mentioned in this file.

tasks/centos_deploy.yml
blemagicleap: To support or install pip(3) on CentOS 7 I must run /usr/bin/easy_install-3.4 command. However, I can't replace the python version release (4) to make it works on python 2. Any suggestions?

blemagicleap: Bty, i already fixed the idempotence issue.

templates/configuration.py.j2
blemagicleap: Removed the templates directory.

tasks/centos_deploy.yml
blemagicleap: That's right!

templates/configuration.py.j2
lae: You should push your changes and your commit messages should say this. I don't see it removed and it's been 2 days.

Issue #4, opened 2017-05-04
Template standard netbox configuration


lae:
Configuration files should be stored in `netbox_shared_path` and symlinked into the appropriate deployment directory (stable/tarball release or git directory).

[File to be templated](https://github.com/digitalocean/netbox/blob/develop/netbox/netbox/configuration.example.py)

Only the relevant fields should be configurable. Some should use already defined role variables.

http://netbox.readthedocs.io/en/stable/installation/netbox/#configuration
http://netbox.readthedocs.io/en/stable/configuration/mandatory-settings/
http://netbox.readthedocs.io/en/stable/configuration/optional-settings/

PR #9, opened 2017-05-24
Configure Netbox and perform deployment tasks

blemagicleap:
Please review the code and give me some suggestion to improve coding.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,3 +25,16 @@ netbox_database_port: 5432
 
 netbox_python: 3
 
+# netbox_secret_key:  "Generate the secrec_key if defined"
+netbox_secret_key:  "V1kp=u-LvT^H_*IUa!J$8ElzS&sdMD#etw2qP6gcARi94)r7Wf"
+
+netbox_allowed_hosts: 
+ - netbox
+ - localhost
+
+netbox_login_required: false
+netbox_base_path:
+netbox_paginate_count: 50
+netbox_timezone: UTC
+netbox_prefer_ipv4: false
+
--- a/tasks/configuration.yml
+++ b/tasks/configuration.yml
@@ -1,0 +1,34 @@
+---
+- name: Configure netbox
+  template:
+    src: templates/configuration.py.j2
+    dest: "{{ netbox_shared_path }}/configuration.py"
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+    mode: 0640
+
+- name: Create a link for configuration.py
+  file: 
+    src: "{{ netbox_shared_path }}/configuration.py"
+    dest: "{{ netbox_config_path }}/configuration.py"
+    state: link
+
+- name: Run DB migrations
+  django_manage:
+    command: migrate
+    app_path: "{{ netbox_current_path }}/netbox"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+  become_user: "{{ netbox_database_user }}"
+
+- name: Collect Static Files
+  django_manage:
+    command: collectstatic --no-input
+    app_path: "{{ netbox_current_path }}/netbox"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+
+- name: Load initial Data (optional)
+  django_manage:
+    command: loaddata initial_data
+    app_path: "{{ netbox_current_path }}/netbox"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+  become_user: "{{ netbox_database_user }}"
--- a/templates/configuration.py.j2
+++ b/templates/configuration.py.j2
@@ -1,0 +1,24 @@
+ALLOWED_HOSTS =  {{ netbox_allowed_hosts | to_json }}
+
+DATABASE = {
+    'NAME': '{{ netbox_database }}',                    # Database name
+    'USER': '{{ netbox_database_user }}',               # PostgreSQL username
+{% if netbox_database_host is defined %}
+    'PASSWORD': '{{ netbox_database_password }}',       # PostgreSQL password
+    'HOST':  '{{ netbox_database_host }}',              # Database server
+    'PORT': '{{ netbox_database_port }}',               # Database port (leave blank for default)
+{% endif %}
+}
+
+SECRET_KEY = '{{ netbox_secret_key }}'
+
+LOGIN_REQUIRED = {{ "True" if netbox_login_required else "False" }}
+
+BASE_PATH = '{{ netbox_current_path}}/netbox'
+         
+PAGINATE_COUNT = '{{ netbox_paginate_count }}'
+         
+TIME_ZONE = '{{ netbox_timezone }}'
+
+PREFER_IPV4 = '{{ "True" if netbox_prefer_ipv4 else "False" }}'
+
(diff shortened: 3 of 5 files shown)


defaults/main.yml
```diff
@@ -21,6 +21,25 @@ netbox_database_user: netbox
 #netbox_database_password: changeme
 #netbox_database_host: localhost
 netbox_database_port: 5432
-#netbox_database_socket: /var/run/postgresql
+netbox_database_socket: /var/run/postgresql
```
lae: You're selecting a postgresql connection scenario here by default again. I left it commented out to let the user pick if they want to use a local postgresql server or an external postgresql server, which we check for when we go to create the database later.

Uncommenting it here causes a conflict when someone tries to configure a TCP connection.

tasks/centos_deploy.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Install pip via easy_install
+  command: /usr/bin/easy_install-{{ netbox_python }}.4 pip creates=/usr/bin/pip
```
lae: This just looks bad. I already removed this task entirely from your previous PR, but the issue here is when `netbox_python` is set to `2`, this task doesn't make sense. No one uses Python 2.4 anymore.

Also, you can use the `args` param to make the syntax more readable:

```
  command: /usr/bin/easy_install-3.4 pip
  args:
    creates: /usr/bin/pip 
```

tasks/configuration.yml
```diff
@@ -0,0 +1,33 @@
...
+  django_manage:
+    command: loaddata initial_data
+    app_path: "{{ netbox_current_path }}/netbox"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+
+- name: Test the Application (optional)
+  django_manage:
+    command: runserver 0.0.0.0:8000 --insecure
+    app_path: "{{ netbox_current_path }}/netbox"
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
```
lae: This is unnecessary for a production deployment?

vars/main.yml
```diff
@@ -1,2 +1,27 @@
...
+netbox_ldap_group_cache_timeout: 3600
+netbox_ldap_group_flags:
+   is_active: cn=active,ou=groups,dc=example,dc=com
+netbox_ldap_attr_map:
+   first_name: givenName
+   last_name: sn
+
+# One or the other:
+netbox_ldap_user_filter: (uid=%(user)s)
+#netbox_ldap_user_dn_template: uid=%(user)s,ou=users,dc=example,dc=com
```
lae: So I'm not quite sure why these are all defined here - having the variables here in `vars/` prevents users from setting these themselves. But also, enabled by default? I would assume most people would use local authentication instead of LDAP authentication.

It doesn't look like you're using these variables anywhere in this PR, however...

vars/main.yml
```diff
@@ -1,2 +1,27 @@
 ---
 # vars file for lae.netbox
+
+netbox_app_path: "{{ netbox_current_path }}/netbox/netbox"
```
lae: I don't see this variable being used anywhere.

templates/configuration.py.j2
```diff
@@ -0,0 +1,31 @@
+ALLOWED_HOSTS =  {{ netbox_allowed_hosts | to_json }} 
+
+DATABASE = {
+    'NAME': '{{ netbox_database }}',    	     	# Database name
+    'USER': '{{ netbox_database_user }}',               # PostgreSQL username
+{% if netbox_database_host is defined %}
+    'PASSWORD': '{{ netbox_database_password }}',       # PostgreSQL password
+    'HOST':  '{{ netbox_database_host }}',     		# Database server
+    'PORT': '{{ netbox_database_port }}',               # Database port (leave blank for default)
+{% endif %}
```
lae: Based on my message from Slack you'll want to set `HOST` to the unix domain socket when not using TCP.

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Read existing secret key
+  shell: >
+    cat {{ netbox_current_path }}/netbox/netbox/configuration.py
+    | sed -n "s/^SECRET_KEY *= *'\(.*\)'/\1/p"
+  register: __existing_key
+  failed_when: false
+  ignore_errors: true
```
lae: `failed_when` and `ignore_errors` is redundant, I think...

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Read existing secret key
+  shell: >
+    cat {{ netbox_current_path }}/netbox/netbox/configuration.py
+    | sed -n "s/^SECRET_KEY *= *'\(.*\)'/\1/p"
+  register: __existing_key
```
lae: You should still namespace this with `netbox`, e.g. `__netbox_existing_key`

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
...
+  failed_when: false
+  ignore_errors: true
+  changed_when: __existing_key.rc != 0 or __existing_key.stdout|length == 0
```
lae: This breaks idempotency checking unnecessarily.

defaults/main.yml
```diff
@@ -21,6 +21,25 @@ netbox_database_user: netbox
...
+netbox_paginate_count: 50
+netbox_timezone: UTC
+netbox_prefer_ipv4: false
+
+netbox_date_format: "N j, Y"
+netbox_short_date_format: "Y-m-d"
+netbox_time_format: "g:i a"
+netbox_short_time_format: "H:i:s"
+netbox_datetime_format: "N j, Y g:i a"
+netbox_short_datetime_format: "Y-m-d H:i"
```
lae: It's very overwhelming to have several of these date formatting variables as single variables. Since they all almost serve similar purposes, it would make sense to have this as some sort of dictionary like:

```
netbox_config_date_formats:
  date: "N j, Y"
  short_date: "Y-md"
```

But really, the likelihood of modifying this is very low. I tend to not make these configurable unless there's a need for it, precisely because having too many variables can be daunting....

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
+---
```
lae: This file isn't included anywhere, are you using it?

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
...
+  changed_when: __existing_key.rc != 0 or __existing_key.stdout|length == 0
+
+- name: Use existing secret key
+  set_fact:
+    netbox_secret_key: "{{ __existing_key.stdout }}"
+  when: not __existing_key.changed
+
+- name: Generate secret key
+  set_fact:
+    netbox_secret_key: "{{ lookup('password', '/dev/null length=64 chars=ascii_letters,digits') }}"
```
lae: If someone sets `netbox_secret_key` theirselves, they would run into unexpected behaviour here because you're overriding their definition (if this file were included that is).

tasks/secret_key.yml
```diff
@@ -0,0 +1,19 @@
...
+  when: __existing_key.changed
```
lae: Interesting experiment but, it would probably be best if you just enforced that the user set `netbox_secret_key` themselves in their playbook instead of trying to generate it within the role (and I guess provide instructions for generating one, perhaps).

lae:
You can force push your recreated branch, or we can just close this PR and have another PR with your recreated branch.

lae:
Actually, please re-read #4 as well. The current changeset does not do what the issue requested.

defaults/main.yml
blemagicleap: I'm removing these variables (date_formats) from defaults/main.yml and templates/configuration.py.j2.

Issue #6, opened 2017-05-05
Install/configure gunicorn and service


lae:
Should be installed within the virtualenv for the active deployment. Template out the configuration, store in `netbox_shared_path`.

Preferably, to be less intrusive, this should also be configured as a service using systemd unit files/sysvinit scripts (check using Ansible facts) rather than supervisord.

Once this is done, a test should be added to `.travis.yml` to check that the gunicorn server is running and returns an HTTP 200. See [lae.proxmox](https://github.com/lae/ansible-role-proxmox/blob/master/.travis.yml#L31) for an example.

http://netbox.readthedocs.io/en/stable/installation/web-server/#gunicorn-installation

lae:
Resolved in #13.

PR #10, opened 2017-05-24
Feature/gunicorn supervisord

blemagicleap:
I also finished the gunicorn and supervisor issue. Please review it.


lae:
#6 says to use a systemd unit file...

While the NetBox documentation provides instructions to install supervisord and configure it that way, it introduces a point of failure between the init system and the application. There's no need to install yet another application here. Can you rewrite this to not use supervisord?

lae:
for future reference this was replaced by #12

PR #12, opened 2017-05-31
Git deployment and Gunicorn service

blemagicleap:
Hi Musee,
Per your requested, I replaced a supervisor/d service with systemd. Please reviewing the codes and let me know if you need any changes.

Thanks,
Ben


tasks/gunicorn_config.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install gunicorn
+  pip:
+    name: gunicorn
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
+
+- name: Write gunicorn config
+  template:
+    src: templates/gunicorn_config.py.j2
```
lae: Don't prefix this with `templates/` - the `template` module already looks in the `templates` directory by default.

tasks/gunicorn_config.yml
```diff
@@ -0,0 +1,24 @@
...
+    dest: "{{ netbox_current_path }}/netbox/gunicorn_config.py"
+
+- name: Write Netbox service for gunicorn
+  template:
+    src: templates/netbox.service
```
lae: remove `templates` prefix

vars/main.yml
```diff
@@ -2,3 +2,5 @@
 # vars file for lae.netbox
 
 netbox_config_path: "{{ netbox_current_path }}/netbox/netbox"
+netbox_venv_path: "{{ netbox_current_path }}//venv-py{{ netbox_python }}/bin"
```
lae: Variable name implies this is the path to the virtualenv, but you're giving the path to the /bin directory in the virtualenv.

I suggest removing `/bin` and using this variable everywhere as specified in other comments.

vars/main.yml
```diff
@@ -2,3 +2,5 @@
...
+netbox_gunicorn_port: 8001
```
lae: This makes sense to leave configurable by the user, and so should be in `defaults`

tasks/gunicorn_config.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install gunicorn
+  pip:
+    name: gunicorn
+    virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
```
lae: Could use `netbox_venv_path` here.

tasks/gunicorn_config.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: Write gunicorn config
+  template:
+    src: templates/gunicorn_config.py.j2
+    dest: "{{ netbox_current_path }}/netbox/gunicorn_config.py"
```
lae: Config file is supposed to be stored in the `shared` directory, specified by issue #6

tasks/gunicorn_config.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: Write Netbox service for gunicorn
+  template:
+    src: templates/netbox.service
+    dest: /lib/systemd/system
+    owner: root
+    mode: 0644
+  notify: restart netbox.service
```
lae: Changes in this task should also trigger a `daemon-reload` in systemd.

tasks/main.yml
```diff
@@ -39,15 +39,13 @@
...
     - "{{ netbox_git_path }}"
 
-- include: download_netbox_tarball.yml
-  when: netbox_stable == true
+- include: download_netbox.yml
 
 - name: Install Netbox python dependencies inside active virtualenv
   pip:
     requirements: "{{ netbox_current_path }}/requirements.txt"
     virtualenv: "{{ netbox_current_path }}/venv-py{{ netbox_python }}"
```
lae: Replace with `netbox_venv_path` probably.

tasks/main.yml
```diff
@@ -75,4 +73,7 @@
 
 - name: Postgres template and configuration 
   include: configuration.yml
-  when: netbox_stable
+
+- name: Configure gunicorn
```
lae: includes don't need to have names, they're not tasks by theirselves, just helpers.

templates/gunicorn_config.py.j2
```diff
@@ -0,0 +1,6 @@
+command = '{{ netbox_venv_path }}/gunicorn'
+pythonpath = '{{ netbox_current_path }}/netbox'
+bind = '127.0.0.1:{{ netbox_gunicorn_port }}'
```
lae: bind address should be fully configurable in my opinion.

PR #13, opened 2017-06-07
NetBox deployment polish

### LINKED ISSUE #3: Deploy netbox via git
Deployment and configuration should follow same pattern as #1, not much should differ, except in the cases where configuration files might change. For release we'll focus on current `develop`.

- `netbox_git_path` should store the git repository - unlike the stable scenario (where there would be multiple folders for different releases).
- `netbox_current_path` should be symlinked to `netbox_git_path`.
- Changes to git repository should trigger a restart/reload.

This should support users who want to use their own forks as well (hence the need for the `netbox_git_uri` variable, which you can already see in defaults).

http://netbox.readthedocs.io/en/stable/installation/netbox/#option-b-clone-the-git-repository

lae:
Main changes:

- test that NetBox does successfully deploy with an HTTP request
- perform all tasks regarding deployment of the netbox application and its dependencies as the user specified in `netbox_user`
- don't specify a SECRET_KEY by default (doing this opens people who don't set it to something different theirselves open to attack) and generate one using the provided script if users don't configure one
- fix the `django_manage` tasks that were being skipped/not executed correctly

Also completes the remaining missing requirements from and closes #3 and #6.


lae:
actually, this is missing reload actions when configuration/deployment changes

PR #27, opened 2018-05-29
NetBox 2.4 release preparation

lae:
This branch currently is not ready for production until NetBox 2.4 is released, and is intended to track changes upstream that might break deployment (there is a travis cronjob configured to test this branch daily).

You can use this branch to deploy the NetBox 2.4 branch as usual by installing it using `ansible-galaxy` like below.

```
ansible-galaxy install https://github.com/lae/ansible-role-netbox/archive/feature/netbox-2.4.tar.gz,v0.6.0-pre,lae.netbox --force
```

You must specify the following role variables to test the 2.4 branch:

```
netbox_git: true
netbox_git_version: develop-2.4
```

## Changes

* Python 2 support is dropped from the role (still supported by NetBox 2.4 but will be dropped in 2.5 - I'd rather not keep Python 2 support around if no one is configuring a Python 2 deployment with this role) - #26 
* Workarounds to fix bugs/shortcomings of NetBox <2.3 have been removed - the next version, v0.6.0 should only be used to deploy NetBox 2.3 or higher. Users should stick to v0.5.6 otherwise.
* Configuration files are now validated - prevents services from breaking - #18 
* Ubuntu 18 LTS is now supported
* Future-proofing changes for Ansible 2.6 (and one for 2.10) - minimum required Ansible version is now 2.4 
* Role is now tested with Python 3 and fixes were made to support Python 3

If you'd like to see anything added, deployment-wise, in the next release, feel free to open an issue.

PR #29, opened 2018-07-22
Add Ubuntu 18.04 Bionic support

logan2211:


lae:
Hi @logan2211! Thanks for the PR, but Ubuntu 18 support has actually been implemented in #27 and this will conflict with changes in that branch (specifically because Python 2 support is dropped), so I'm closing this as a duplicate.

If you're in a rush to use Ubuntu 18, it should be fine to use that branch (feature/netbox-2.4) as-is for the stable deployment scenario (if you're using git, then you'll need to override `netbox_git_version`), since it'll eventually be merged as-is once NetBox 2.4 hits.

logan2211:
Thanks! Sorry for the noise! I did not see #27.

By the way, really neat testing setup with the LXC profiles. 👍

PR #30, opened 2018-08-01
Add support for Ubuntu Bionic (18.04)

mnaser:
This patch adds support for Ubuntu Bionic (18.04) alongside with
all the changes required in order to add testing in Travis CI.

Due to the fact that Ubuntu Bionic only ships with Python 3.7 which
is not supported by Django 1.x which Netbox uses, this patch makes
sure that it is only deployed using Python 2.7.


mnaser:
Wow, so I submitted a patch by Logan and I know him in person, that's quite hilarious.

lae:
Hi, based on your last message I think you've already seen #29, but in case you haven't, please see my comment there. I'm closing this as a dupe.

PR #31, opened 2018-08-10
Suppot RHEL and add example playbook

as0bu:
Running this example on a RHEL installs EPEL. Use at your own risk


lae:
Hi @as0bu, thanks for the pull request.

I notice that `vars/redhat-7.yml` is not any different from `vars/centos-7.yml`. Since RedHat would be the OS family for both CentOS and RedHat, could I have you either amend your commit or push a commit to remove the CentOS vars file so that there's no unneeded duplication?

Other than that, this PR looks fine. Test build errors seem to be transient. I'd restart them, but will wait for the above change.

as0bu:
I need to fix one other thing as well. The default PostgreSQL version in Red Hat 7 is too old for jsonb. So I'll be updating the example I provided in an earlier commit.

lae:
The other example uses a separate role to fulfill the 9.3 postgres dependency, actually. See https://github.com/lae/ansible-role-netbox/blob/master/examples/playbook_single_host_deploy.yml#L5 (and line 17)

While I don't mind having the method you have in the example, what are your thoughts? Keep as-is?

as0bu:
@lae I think we should keep that as is. The problem is that the default repos for Red Hat 7 only have PostgreSQL version 9.2, unless you enable SCL which I don't think will work for this. I think for anything other than Red Hat, that example works as expected.

PR #34, opened 2018-09-22
check universe repo for ubuntu server in install_packages_apt.yml

Side2005:
check if ubuntu universe is available


lae:
@Side2005 I can't recall any discussion about this and I'm not sure why I merged this. What was the reasoning behind this PR? I don't recall any breakage prior to adding this.

PR #38, opened 2018-11-08
Implement webhooks support

### LINKED ISSUE #37: Add option to enable webhooks
Webhooks are not enabled by default with netbox install so additional packages and configuration is required
(as per https://netbox.readthedocs.io/en/latest/installation/2-netbox/#webhooks-configuration)

install redis
install django-rq via pip

addition configuration required
```
WEBHOOKS_ENABLED = True
REDIS = {
    'HOST': 'localhost',
    'PORT': 6379,
    'PASSWORD': '',
    'DATABASE': 0,
    'DEFAULT_TIMEOUT': 300,
}
```

madeinoz67:
This implements webhooks option and will install dependencies required during install

closes #37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_script:
 - echo "host all all 10.0.3.0/24 trust" | sudo tee -a /etc/postgresql/9.4/main/pg_hba.conf
 - echo "listen_addresses = '*'" | sudo tee -a /etc/postgresql/9.4/main/postgresql.conf
 - sudo service postgresql restart
+- sudo redis-server /etc/redis/redis.conf --bind 10.0.3.1 --protected-mode no --port 6379
 - cd tests/
 script:
 - ansible-playbook -i inventory deploy.yml --syntax-check
--- a/README.md
+++ b/README.md
@@ -56,6 +56,10 @@ separate from this role. Take a look at the *Example Playbook* section.
 
 Please note that as of v2.2, NetBox requires PostgreSQL 9.4 or higher.
 
+### Redis 
+
+If you require webhooks then a redis server instance is required.  This role does not setup or manage the redis instance, see galaxy role [DavidWittman.redis](https://galaxy.ansible.com/davidwittman/redis)
+
 Role Variables
 --------------
 
@@ -183,6 +187,16 @@ You can find an example in `examples/`.
 Toggle this to enable NAPALM integration in NetBox. You must define
 `NAPALM_USERNAME` and `NAPALM_PASSWORD` in the `netbox_config` variable to be
 able to use NAPALM.
+
+    netbox_webhooks_enabled: false
+    netbox_redis_host: 127.0.0.1
+    netbox_redis_port: 6379
+    netbox_redis_password: ''
+    netbox_redis_database: 0
+    netbox_redis_default_timeout: 300
+
+Toggle `netbox_webhooks_enabled` to `true` to enable webhooks for
+NetBox.  REDIS server instance is required to use webhooks (see prerequisites)
 
 Example Playbook
 ----------------
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -21,6 +21,14 @@ netbox_database_user: netbox
 netbox_database_port: 5432
 #netbox_database_socket: /var/run/postgresql
 
+netbox_webhooks_enabled: false
+netbox_redis_host: 127.0.0.1
+netbox_redis_port: 6379
+netbox_redis_password: ''
+netbox_redis_database: 0
+netbox_redis_default_timeout: 300
+
+
 netbox_config:
   #SECRET_KEY:
   ALLOWED_HOSTS:
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -33,6 +33,17 @@
   become: True
   become_user: "{{ netbox_user }}"
 
+- name: Install django-rq if webhooks enabled
+  pip:
+    name: "django-rq"
+    virtualenv: "{{ netbox_virtualenv_path }}"
+    virtualenv_python: "{{ netbox_python_binary }}"
+  become: True
+  become_user: "{{ netbox_user }}"
+  when:
+    - netbox_webhooks_enabled
+
+
 - name: Generate NetBox configuration file
   template:
     src: templates/configuration.py.j2
--- a/templates/configuration.py.j2
+++ b/templates/configuration.py.j2
@@ -11,6 +11,17 @@ DATABASE = {
 {% endif %}
 }
 
+{% if netbox_webhooks_enabled %}
+WEBHOOKS_ENABLED = True
+REDIS = { 
+    'HOST': '{{ netbox_redis_host }}',
+    'PORT': '{{ netbox_redis_port }}',
+    'PASSWORD': '{{ netbox_redis_password }}',
+    'DATABASE': '{{ netbox_redis_database }}',
+    'DEFAULT_TIMEOUT': '{{ netbox_redis_default_timeout }}',
+}
+{% endif %}
+
 {% for setting, value in netbox_config.items() %}
 {% if value in [True, False] %}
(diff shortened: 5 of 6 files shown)


madeinoz67:
I have all tests passing except Ubuntu-xenial if someone could please have a look to see what I'm doing wrong as I'm a centos guy and not 100% on ubuntu.  I thinks it may be around redis-server package.

lae:
Hey Stephen, first, thanks for the PR!

However, I cannot merge this PR as-is, mainly because it expands the scope of this role heavily.

1. Redis shouldn't be installed by this role. Ideally, Redis is managed by a separate Ansible role, i.e. users are free to use their own method of setting up Redis (perhaps they want HA? for example). If you note, this role doesn't install Postgres either.
  a. I was looking up roles earlier that could be good contenders, possibly https://github.com/DavidWittman/ansible-redis?
  b. For testing in Travis, we should use the one provided by the build image: https://docs.travis-ci.com/user/database-setup/#redis (sounds like the bind address must be corrected however)
2. Ideally, just toggling `netbox_webhooks_enabled` should be enough, in my opinion, to enable webhook support with defaults.
  a. We could inject `netbox_config` with options, but I think this can be done similarly to how `netbox_database` and friends are handled.
3. I just realised/remembered that I accidentally committed a duplicate tasks file - I just fixed this in a6240992ab13485ae8727b451cd5933b26b7023f. Could you rebase, as well as squash your commits into fewer commits and then force push to this branch? (to avoid excessive pollution of commit history)

madeinoz67:
Makes sense @lae.  I'll look at doing knocking these off tomorrow as is late here.

lae:
Alright, thanks! P.S. if you need any help, I just added a Discord link to the README you can use to reach me.

lae:
Hi @madeinoz67, just checking up on this. Have you made any progress?

madeinoz67:
I've just made the changes requested and testing the with travis redis server now, then will rebase and squash

.travis.yml
```diff
@@ -29,6 +29,7 @@ before_script:
 - echo "host all all 10.0.3.0/24 trust" | sudo tee -a /etc/postgresql/9.4/main/pg_hba.conf
 - echo "listen_addresses = '*'" | sudo tee -a /etc/postgresql/9.4/main/postgresql.conf
 - sudo service postgresql restart
+- sudo redis-server /etc/redis/redis.conf --port 6379 --requirepass ''
```
lae: Gonna wait for the tests from your most recent commit to run but I think you might need to specify binding to 0.0.0.0 (since I think it probably binds to localhost by default).

tasks/deploy_netbox.yml
```diff
@@ -33,6 +33,18 @@
   become: True
   become_user: "{{ netbox_user }}"
 
+- block:
```
lae: can you move this out of a block since it's only one task?

README.md
```diff
@@ -178,6 +185,19 @@ Toggle this to enable NAPALM integration in NetBox. You must define
...
+Toggle `netbox_webhooks_enabled` to `true` to enable webhooks for
+NetBox, however you will required redis server installed first. You also need to add the following to the `netbox_config:`
+
+      WEBHOOKS_ENABLED: true
+      REDIS: 
+        'HOST': 'localhost'
+        'PORT': 6379
+        'PASSWORD': ''
+        'DATABASE': 0
+        'DEFAULT_TIMEOUT': 300
```
lae: As I was talking about in point 2a from my last comment, I think this could be better if we had `netbox_webhooks_enabled` as you already have, and then sane defaults for Redis configuration as separate role variables outside of `netbox_config` (e.g. `netbox_redis_host`) which are then injected into the configuration template separately. This would make it possible for most people (who'd install redis locally) to enable webhook support with just one role variable and just specify any that differ from defaults. What do you think?

README.md
```diff
@@ -178,6 +185,19 @@ Toggle this to enable NAPALM integration in NetBox. You must define
 `NAPALM_USERNAME` and `NAPALM_PASSWORD` in the `netbox_config` variable to be
 able to use NAPALM.
 
+    netbox_webhooks_enabled: false
+
+Toggle `netbox_webhooks_enabled` to `true` to enable webhooks for
+NetBox, however you will required redis server installed first. You also need to add the following to the `netbox_config:`
```
lae: I'd like to mention at least one role for installing Redis here, but I haven't gotten around to testing the one I mentioned before. Maybe later, unless you've tried one already?

.travis.yml
lae: ignore this, I think you got it now actually

tasks/deploy_netbox.yml
madeinoz67: will do

README.md
madeinoz67: I agree, the less config the end user has to think about the better.  I was thinking along these lines as well and was going to look at this once I get everything passing on Travis.

.travis.yml
madeinoz67: ...finally!!!

madeinoz67: ok cooking with gas now!

README.md
madeinoz67: No I haven't tested any role for REDIS as yet, was going to look at the one you suggested and give it a spin.

tasks/deploy_netbox.yml
madeinoz67: done

README.md
madeinoz67: done

templates/configuration.py.j2
```diff
@@ -11,6 +11,16 @@ DATABASE = {
 {% endif %}
 }
 
+{% if netbox_webhooks_enabled %}
+ WEBHOOKS_ENABLED: true
```
lae: this is a python file, syntax is different, you need to use `=`

basically this entire section needs to look like what's here https://netbox.readthedocs.io/en/latest/installation/2-netbox/#webhooks-configuration

(also booleans are case-sensitive)

madeinoz67: ...so I found out lol 
all fixed and passing

README.md
madeinoz67: just tested the role mentioned and will include this in the updated README

madeinoz67: updated README prerequisites and webhook variables, if you're happy then this is resolved

defaults/main.yml
```diff
@@ -21,6 +21,14 @@ netbox_database_user: netbox
 netbox_database_port: 5432
 #netbox_database_socket: /var/run/postgresql
 
+netbox_webhooks_enabled: false
+netbox_webhooks_redis_host: 127.0.0.1
```
lae: Could we change these variables to not have the `webhooks` namespace and just be `netbox_redis_X`? It's possible that NetBox might use Redis for something else than just webhooks in the future, which wouldn't require a separate Redis definition, I think?

defaults/main.yml
```diff
@@ -30,6 +38,13 @@ netbox_config:
   #NAPALM_PASSWORD:
   MEDIA_ROOT: "{{ netbox_shared_path }}/media"
   REPORTS_ROOT: "{{ netbox_shared_path }}/reports"
+  #WEBHOOKS_ENABLED: true
+  #REDIS: 
+  #  'HOST': 'localhost'
+  #  'PORT': 6379
+  #  'PASSWORD': ''
+  #  'DATABASE': 0
+  #  'DEFAULT_TIMEOUT': 300
```
lae: needs to be removed

lae:
Okay, apart from my last two comments, looks good. When you're done, can you squash everything into one descriptive commit and force push to this branch? (if you need help figuring this step out, let me know)

defaults/main.yml
madeinoz67: good pickup, done

defaults/main.yml
madeinoz67: sure thing

madeinoz67:
squashed and rebased over to you...

PR #43, opened 2018-11-14
Add tests using molecule with vagrant

madeinoz67:
Adds vagrant molecule to make locally testing with vagrant much easier and cleaner

Requires molecule to be installed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,161 @@
+# File created using '.gitignore Generator' for Visual Studio Code: https://bit.ly/vscode-gig
+
+# Created by https://www.gitignore.io/api/python,visualstudiocode,vagrant
+# Edit at https://www.gitignore.io/?templates=python,visualstudiocode,vagrant
+
+### Python ###
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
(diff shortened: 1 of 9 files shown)


lae:
I think builds failed because upstream pushed a commit right in between the deploy and idempotency test. Lucky.

madeinoz67:
Forgot to mention that Ive disabled lint checks in this PR as was throwing up a number or errors throughout the code which is out of scope of the PR to fix them.  

Lint tests can be re-enabled in molecule.yml under their relevant sections
e.g.

```yaml
lint:
  name: yamllint
  enabled: true
```

```yaml
provisioner:
  name: ansible
  log: true
  lint:
    name: ansible-lint
    enabled: true
```

lae:
* Can you add a `DEVELOPING.md` that discusses briefly how to setup and run this test environment?
* The `.gitignore` is pretty unwieldy. What files do you actually have in your dev environment that you need to ignore? `.vscode/*`? (afaict the only lines relevant to this repo are `.vagrant`).
* Does `molecule/` need to be a top level directory or does it make sense to move it into the `tests/` directory?

madeinoz67:
@lae can do not a problem.

I develop in vscode and its part of my standard .gitignore so I don't commit anything I shouldn't but can be removed.

molecule needs to be as it is currently set as when running tests you run from the root of the role.

e.g. molecule test

All tests reside in the molecule/scenarios/test, in this case a single "default" scenario. 

Other scenarios can be added at anytime in the future using molecule init scenario .....

madeinoz67:
@lae please review DEVELOPING.md to see it its what you want

lae:
Commits were cherry-picked and merged.

PR #46, opened 2019-03-06
napalm: allow using custom package list

mnaser:
This change will allow the deployer to use a custom NAPALM package
list which will allow them to use community drivers for devices
that are not natively supported via NAPALM (such as RouterOS).


lae:
Hi, thanks for the PR. Could you update the Readme with this role variable as well?

lae:
Documentation: https://github.com/lae/ansible-role-netbox/commit/e09129c37800e4434cc24c456a16d74c4250f374

Issue #48, opened 2019-03-22
Method for skipping EPEL install step


tyler-8:
[This task](https://github.com/lae/ansible-role-netbox/blob/master/tasks/install_packages_yum.yml#L2) should be made optional via a variable or tag. In many enterprise RHEL environments, installing your own repository can be a big no-no, or an EPEL clone for the environment may already exist that isn't going to match the public EPEL package in yum.

For instance:
`netbox_install_epel = false`

```
- name: Install EPEL repository
  yum:
    name: epel-release
    state: installed
  when: netbox_install_epel
```

PR #49, opened 2019-03-22
Add var to skip epel install

tyler-8:
Fixes https://github.com/lae/ansible-role-netbox/issues/48


tyler-8:
> Also, it should be noted that if this variable is toggled, the end user is responsible for providing the correct Python packages for NetBox.

Assuming this is being disabled because the user's enterprise RHEL environment has an internal EPEL (rather than installing EPEL direct from FedoraProject), it shouldn't cause any additional steps or work.

PR #50, opened 2019-05-02
Allow role to run successfully in --check mode

sigio:
In check-mode, the register task for _path_exec isn't executed, which then fails the set_fact task after it.
Since these are safe to run in check-mode, i've added the check_mode:false flag.


sigio:
Thanks for the very quick merge ;)

lae:
@sigio no problem lol

PR #53, opened 2019-05-17
Support deeply nested configurations

0x5a17ed:
A configuration value can become deeply nested especially 
when dealing with `NAPALM_ARGS`.  In this scenario the 
approach to dump the configuration tree with `to_nice_json`
can lead to the creation of a configuration file that is not
valid Python anymore.  Store these values as strings in the
`configuration.py` file and load them with `json.loads`.

### DIFF
--- a/templates/configuration.py.j2
+++ b/templates/configuration.py.j2
@@ -1,3 +1,5 @@
+import json
+
 {{ ansible_managed | comment }}
 DATABASE = {
     'NAME': '{{ netbox_database }}',
@@ -13,7 +15,7 @@ DATABASE = {
 
 {% if netbox_webhooks_enabled %}
 WEBHOOKS_ENABLED = True
-REDIS = { 
+REDIS = {
     'HOST': '{{ netbox_redis_host }}',
     'PORT': '{{ netbox_redis_port }}',
     'PASSWORD': '{{ netbox_redis_password }}',
@@ -25,8 +27,10 @@ REDIS = {
 {% for setting, value in netbox_config.items() %}
 {% if value in [True, False] %}
 {{ setting }} = {{ 'True' if value else 'False' }}
-{% else %}
+{% elif value is string or value is number %}
 {{ setting }} = {{ value | to_nice_json }}
+{% else %}
+{{ setting }} = json.loads(r'''{{ value | to_json }}''')
 {% endif %}
 {% endfor %}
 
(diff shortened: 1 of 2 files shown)


tests/install.yml
```diff
@@ -12,3 +12,4 @@
     test_host_suffixes:
       - stable
       - git
+      - napalm
```
lae: It's not quite obvious but this significantly increases test time and disk usage within the test VM (which could lead to out of space errors in CI). NAPALM setup is not a significantly different scenario to justify bringing up a whole set of different containers I think - the group vars should just be applied to all of them (in `group_vars/netbox` as stated earlier).

0x5a17ed: Alright, I'll take care of this tomorrow. Thanks for the input :+1:

0x5a17ed: I didn't know that adding a new profile would have this kind of consequences. My initial intention was to add a second test case with NAPALM enabled since it looked like a good idea to me to test both paths: one with NAPALM enabled and one with NAPALM disabled.

0x5a17ed:
@lae do you have an idea why the CI pipeline is failing now? It seems to struggle with installing the napalm requirement. Maybe I overlooked the cause but it doesn't seem obvious to me.

lae:
@0x5a17ed one of the dependencies apparently broke https://github.com/ncclient/ncclient/issues/321

They pushed out a fix and called 0.6.5 back so I'm going to rerun the test.

(including error for posterity):

```
Collecting ncclient>=0.5.4 (from junos-eznc>=2.2.0->napalm)
  Using cached https://files.pythonhosted.org/packages/d2/1f/4333dba7d63bebb27232d0cd5d2173ed14615f50c92a41479b86aab3217d/ncclient-0.6.5.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-j0vzfzw7/ncclient/setup.py", line 37, in <module>
        test_req_lines = [line.strip() for line in open("test-requirements.txt").readlines()]
    FileNotFoundError: [Errno 2] No such file or directory: 'test-requirements.txt'
```

tests/install.yml
lae: Yeah, that's understandable. Anyway, LGTM.

0x5a17ed:
Thank you very much for helping out and taking care of this :hugs:

Issue #51, opened 2019-05-13
Upgrade of psycopg2 dependency fails


madeinoz67:
I've been away on leave and when I returned last week I bumped netbox to the latest version, low and behold its failed to deploy psycopg2 python package, where before its been fine.

It appears that pg_config can't be found  when pip goes to install psycopg2.  Looking back at my previous successful deploys it was using v0.7.6 of this role.  changes seem to indicate that psycopg2 was added in 0.7.7

Ive been trying to track down source of this issue and see that  pg_config is installed on the target machine and available from command shell, yet doesn't appear to be available to pip3?

```
TASK [lae.netbox : Install psycopg2/selinux via pip on Red Hat-based distros] ***
[DEPRECATION WARNING]: Invoking "pip" only once while using a loop via 
squash_actions is deprecated. Instead of using a loop to supply multiple items 
and specifying `name: "{{ item }}"`, please use `name: ['psycopg2', 'selinux']`
 and remove the loop. This feature will be removed in version 2.11. Deprecation
 warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[...]

PR #55, opened 2019-06-27
Feature/netbox 2.6

corny:


madeinoz67:
could we get the workaround for #51 incorporated you look at merging?

README.md
```diff
@@ -191,26 +192,28 @@ define `NAPALM_USERNAME` and `NAPALM_PASSWORD` in the `netbox_config` variable
...
 
-    netbox_keep_uwsgi_updated: false
+    netbox_webhooks_enabled: false
 
 Toggle `netbox_keep_uwsgi_updated` to `true` if you wish to ensure your uwsgi
 server is the latest release, otherwise uwsgi will not be updated on subsequent
 runs of your playbook.
 
-    netbox_install_epel: true
+    netbox_keep_uwsgi_updated: false
```
lae: I'm not sure if this was intentional or not, but the descriptions for role variables comes after the role variable and its default for the majority of the doc. Can you revert these changes/move the redis description below the redis variables?

defaults/main.yml
```diff
@@ -24,10 +24,12 @@ netbox_database_port: 5432
 #netbox_database_socket: /var/run/postgresql
 
 netbox_webhooks_enabled: false
+
 netbox_redis_host: 127.0.0.1
 netbox_redis_port: 6379
 netbox_redis_password: ''
 netbox_redis_database: 0
+netbox_redis_cache_database: 1
```
lae: I don't see this variable being used anywhere - is it necessary?

examples/netbox_config.yml
```diff
@@ -63,6 +68,7 @@ netbox_config:
     PORT: 6379
     PASSWORD: redisisfun
     DATABASE: 0
+    CACHE_DATABASE: 1
```
lae: The `REDIS` section from the config example should be removed entirely, I think, since it's defined by other role variables now.

lae:
Resolved in #61

PR #56, opened 2019-06-27
Use to_nice_json filter

corny:


lae:
This breaks some configs. See #53. Is there some issue that you're having that you're solving with this change?

corny:
Ok, I see. I just tried to simplify the code.

Issue #64, opened 2019-10-22
Provide interface for additional database parameters


tyler-8:
The [configuration template](https://github.com/lae/ansible-role-netbox/blob/master/templates/configuration.py.j2) needs tweaking  in some way. Django (and therefore Netbox) has a number of database specific parameters available that aren't currently able to be passed in easily.

The original requirement that brought this idea up was that [CONN_MAX_AGE](https://docs.djangoproject.com/en/2.2/ref/databases/#persistent-connections) is not set by default in Netbox, but in many cases ought to be. I don't think there's a way to get that parameter passed in currently. Perhaps some sort of for-loop in the Jinja template for any additional keys would suffice. Happy to submit a PR.

lae:
From a brief perusal through the Django documentation, [`OPTIONS`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-OPTIONS) expects a dictionary and I see another parameter that expects a boolean. I'm not sure if it's worth it to support `OPTIONS`, since it seems like actually supporting it fully requires more templating logic (specifically this [isolation level](https://docs.djangoproject.com/en/2.2/ref/databases/#isolation-level) setting). 

The current PR templates everything as strings. Apart from `OPTIONS`, do the integer/boolean settings work as expected while quoted? If not, we could copy the boolean/number logic from the `netbox_config` part of the template.

I wrote the following before I realised about `OPTIONS`, but maybe there's some interest in one of the methods below?

------------

Do you think it might be preferable to have a `netbox_database_config` role variable and get rid of the logic in the templates? i.e. default could be (note, I haven't tested this so it might need modification):

```yaml
netbox_database_config:
  HOST: "{{ netbox_database_socket | default(netbox_database_host  | default('')) }}"
  NAME: "{{ netbox_database }}"
  USER: "{{ netbox_database_user }}"
  PASSWORD: "{{ netbox_database_password if netbox_database_password is defined }}"
  PORT: "{{ netbox_database_port if netbox_database_port is defined }}"
```

Then we could replace the template with:

```python
DATABASE = json.loads(r'''{{ netbox_database_config | to_json }}''')
```

But I guess that would require copying the entire thing if you wanted to add some parameter. We could also just try merging an `extra` config like in your PR with the main config in some task (i.e. create a new private variable) and then dumping it as json in the template.

[...]

PR #65, opened 2019-10-22
Add support for configuring database CONN_MAX_AGE

tyler-8:
Fixes https://github.com/lae/ansible-role-netbox/issues/64

This updates the configuration template and adds documentation for the new `netbox_database_extras` parameter.

### DIFF
--- a/templates/configuration.py.j2
+++ b/templates/configuration.py.j2
@@ -11,9 +11,10 @@ DATABASE = {
 {% else %}
     'HOST': '{{ netbox_database_socket }}',
 {% endif %}
+    'CONN_MAX_AGE': {{ netbox_database_conn_age }},
 }
 
-REDIS = { 
+REDIS = {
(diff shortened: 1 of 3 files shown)


tyler-8:
This should be ready now. I've altered it to only add the ability to pass a value for `CONN_MAX_AGE`

templates/configuration.py.j2
```diff
@@ -11,9 +11,10 @@ DATABASE = {
 {% else %}
     'HOST': '{{ netbox_database_socket }}',
 {% endif %}
+    'CONN_MAX_AGE': '{{ netbox_database_conn_age }}',
```
lae: Templated variable needs to be left unquoted since a number type is expected.

tyler-8: Corrected

tyler-8:
I think I've got it all resolved now

Issue #68, opened 2019-12-24
Add support for Netbox metrics configurations


tyler-8:
When using the built-in metrics for Netbox (based on `django-prometheus`), there are a couple new configuration items that need to be added across the project.

The `systemd` service file would need this parameter set (to auto-create the metrics storage folder in `/run` on boot):

```
[Service]
{% if netbox_metrics_enabled %}
RuntimeDirectory={{ netbox_metrics_dir }}
{% endif %}
```

`uwsgi.ini` - This whole line can be set by the user already using `netbox_uwsgi_options`
```
[uwsgi]
env=prometheus_multiproc_dir=/run/{{ netbox_metrics_dir }}
```

So depending on your preference it may really just be a case of adding the new template snippet to `netbox.service.j2`

lae:
That method looks fine to me. Do you want to open a PR?

tyler-8:
Yes - I'll put one together

PR #69, opened 2020-01-17
Add support for Netbox metrics configurations

tyler-8:
Fixes https://github.com/lae/ansible-role-netbox/issues/68

Adds configuration and templates for metrics configurations.

### DIFF
--- a/tests/test.yml
+++ b/tests/test.yml
@@ -25,3 +25,6 @@
     - name: Ensure that NetBox returns a successful HTTP response
       uri:
         url: "http://{{ inventory_hostname }}:8080"
+    - name: Check that Netbox metrics endpoint works
+      uri:
+        url: "http://{{ inventory_hostname }}:8080/metrics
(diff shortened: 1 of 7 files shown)


tyler-8:
Had to tweak the `uwsgi` template slightly. Ran the full playbook again on a RHEL7 box and all the pieces were there, `/metrics` worked as intended.

lae:
Yeah, I don't doubt it works as intended, but it should still be enabled for the CI tests so that if it does break in any significant manner, it'll be caught.

tyler-8:
Oh my bad, I misunderstood what you said before. Will do!

tests/test.yml
```diff
@@ -25,3 +25,6 @@
     - name: Ensure that NetBox returns a successful HTTP response
       uri:
         url: "http://{{ inventory_hostname }}:8080"
+    - name: Check that Netbox metrics endpoint works
+      uri:
+        url: "http://{{ inventory_hostname }}:8080/metrics
```
lae: Closing quotation mark is missing here.

PR #74, opened 2020-02-15
MVP for 2.7 Install

kdhlab:
This creates the redis DB in the new format for 2.7+ installs. Not heavily tested, I just needed to get something working for my environment and figured I would commit it back if it is helpful.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
 # defaults file for lae.netbox
 netbox_stable: false
-netbox_stable_version: 2.7-beta1
+netbox_stable_version: 2.7.4
 netbox_stable_uri: "https://github.com/digitalocean/netbox/archive/v{{ netbox_stable_version }}.tar.gz"
 
 netbox_git: false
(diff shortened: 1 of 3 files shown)


lae:
NetBox 2.7 is going to result in a version bump from 0.8.x to 0.9.x. I don't think we don't need to support 2.6 for that release (and we didn't do that for 2.6's release when bumping this role from 0.7.x to 0.8.x), since I'm considering that as a major version bump (given it's pre-1.0). Thoughts?

kdhlab:
I'm indifferent. I figured I would write the code to go either way since I am still 2.6 in production, and it really wasn't that much more work.

From the larger perspective of this being a major version bump for this role, have you considered adjusting the supported OSes/python environments as well? I have some internal code that works with CentOS/RHEL 8 that may be worth including.

Maybe open a milestone for .9 separate from the 2.7 support?

lae:
Yeah, I was thinking about RHEL8 support, but I have to update my test suite (which I started working on last week). Those changes would be welcome, too (in fact, there's an issue open about this).

kdhlab:
Closing in preference for a more full featured 2.7 implementation.

PR #32, opened 2018-08-10
Specify netbox packages

as0bu:
This allows users to specify/override the following parameters if needed:

- netbox_packages
- netbox_python_packages
- netbox_python_binary
- netbox_pip_binary
- netbox_ldap_packages

PR #87, opened 2020-04-23
NetBox 2.8 release

lae:
This introduces feature support for NetBox 2.8. If you can test this branch out by upgrading an existing installation (or copy of one), please do so and report back. Confirmations of successful upgrades will result in a quicker merge (as I'm having some test suite issues I need to resolve currently).

To deploy NetBox 2.8, use `ansible-galaxy` to install this branch:

```
ansible-galaxy install https://github.com/lae/ansible-role-netbox/archive/feature/netbox-2.8.tar.gz,v0.10.0-pre,lae.netbox --force
```

## Changes

If I'm missing anything, please leave a comment.
 
- Requirement conditions for Python 3.6+ (complete)
- Plugins/additional Python dependencies support (complete)

Also, I'm considering marking this release as 1.0, since it has generally been stable for a while now. Thoughts?

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -21,10 +21,13 @@
 - name: Ensure the installed Python version is newer than 3.6.0 for NetBox 2.8+
   assert:
     that:
-      - "ansible_python_version | version_compare('3.6.0', '>=')"
-    msg: "NetBox 2.8.0+ requires Python 3.6+. Please either specify an older NetBox version or upgrade to a newer distribution that provides Python 3.6+."
+      - "ansible_python_version is version('3.6.0', '>=')"
+    msg: >
+      NetBox 2.8.0+ requires Python 3.6+. Please either specify an older NetBox
+      version, upgrade to a newer distribution that provides Python 3.6+, or
+      set netbox_python_binary to an appropriate Python 3.6+ binary.
   when:
-    - (netbox_stable and netbox_stable_version | version_compare('2.8.0', '>=') or netbox_git
+    - netbox_stable and netbox_stable_version is version('2.8.0', '>=') or netbox_git
 
 - name: Create NetBox user group
   group:
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -10,6 +10,22 @@
   set_fact:
     ansible_python_interpreter: "{{ netbox_python_binary }}"
 
+- name: Recheck new Python interpreter's version
+  setup:
+    gather_subset: min
+    filter: ansible_python_version
+
+# Can't quite check the effective version of a git installation this early since
+# the repo isn't checked out, but most users will be on develop in the first
+# place so this dumb check should be good enough.
+- name: Ensure the installed Python version is newer than 3.6.0 for NetBox 2.8+
+  assert:
+    that:
+      - "ansible_python_version | version_compare('3.6.0', '>=')"
+    msg: "NetBox 2.8.0+ requires Python 3.6+. Please either specify an older NetBox version or upgrade to a newer distribution that provides Python 3.6+."
+  when:
+    - (netbox_stable and netbox_stable_version | version_compare('2.8.0', '>=') or netbox_git
+
 - name: Create NetBox user group
   group:
     name: "{{ netbox_group }}"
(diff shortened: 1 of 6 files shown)


kdhlab:
For remote authentication from https://github.com/netbox-community/netbox/issues/2328 were you thinking just to allow that to be implemented from the netbox_config variable? That seems to be the simplest solution but I am wondering what will happen if you enable LDAP + that?

lae:
Yeah, from what I've read (release notes and the documentation) it sounds like it can just be managed through the `netbox_config` variable. Do you think this role should do more with regards to it?

I'd expect admins to be aware if they're enabling both - but if there's an issue with enabling both together then we could add an assert task to the `validate_variables.yml` tasks file.

kdhlab:
I just started looking into updating my environment to support it. Right now we do LDAP auth and want to switch to SAML auth which I am sure a lot of people are interested in so I really think it's more of an issue of upgrade paths than anything at least right now. If you upgrade from a 2.7 instance with LDAP turned on to a 2.8 instance what happens to the existing LDAP based users? Will they be picked up in the header based auth as existing users? Will they still be able to login with straight LDAP?

If there is something weird in that whole process, my only thought on how to deal with it is to have either a conversion or user cleanup play that can be called in that scenario to either convert the users in the database or to wipe out the user tables to allow people to start from scratch.

lae:
Ah, I see. This might be a good question to pose on the [netbox-discuss](https://groups.google.com/forum/#!forum/netbox-discuss) mailing list, since Jeremy or someone else might be able to provide advice regarding that.

If it's something we can implement here as an upgrade task, we can consider it, though at the moment we're not tracking what version of NetBox is currently installed, I think, which may or may not pose an issue (maybe not?).

lae:
From another thread:
> I just need a confirmation that upgrading from 2.7 to 2.8 works fine - I may get to this soon (I don't have access to an existing install currently so I need to make one), but if anyone else confirms it, it'll be merged and released.

If you've tried upgrading already with this branch and didn't have any issues, please +1 this post or leave a comment.

kdhlab:
Are we foregoing python version checking?

lae:
Ah, right.

Pushed a commit that I think should work well for existing and new installations.

lae:
The main issue with that is that for new installations, Python 3 may not be installed yet (the role installs them in the appropriate package installation tasks), so we can't really check,

kdhlab:
> The main issue with that is that for new installations, Python 3 may not be installed yet (the role installs them in the appropriate package installation tasks), so we can't really check.

In that case I wouldn't expect netbox_python_binary to be defined yet since we will define that in load_variables.yml if the role is handling the installation. Obviously if we are letting the role handle it we will be setting the variable to a known good one.

lae:
The role does handle it by default, but per #32 the python packages can be overridden by power users (maybe to switch to something like PyPy), so just parsing out the Python versions from the variables specified under `vars/` wouldn't be good enough, I think.

tasks/main.yml
```diff
@@ -10,6 +10,22 @@
...
+# the repo isn't checked out, but most users will be on develop in the first
+# place so this dumb check should be good enough.
+- name: Ensure the installed Python version is newer than 3.6.0 for NetBox 2.8+
+  assert:
+    that:
+      - "ansible_python_version | version_compare('3.6.0', '>=')"
+    msg: "NetBox 2.8.0+ requires Python 3.6+. Please either specify an older NetBox version or upgrade to a newer distribution that provides Python 3.6+."
+  when:
+    - (netbox_stable and netbox_stable_version | version_compare('2.8.0', '>=') or netbox_git
+
```
kdhlab, line 28: This does not work for me on 2.9.7 but the below does:
```suggestion
- name: Ensure the installed Python version is newer than 3.6.0 for NetBox 2.8+
  assert:
    that:
      - "ansible_python_version is version('3.6.0', '>=')"
    msg: "NetBox 2.8.0+ requires Python 3.6+. Please either specify an older NetBox version or upgrade to a newer distribution that provides Python 3.6+."
  when:
    - netbox_stable and netbox_stable_version is version('2.8.0', '>=') or netbox_git
```

Not sure if this is a discrepancy due to ansible versions and the versions test being renamed?

lae: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#version-comparison

Yeah, looks like `version_compare` was renamed in 2.5 and I didn't notice.

lae:
Tested upgrade on Ubuntu 18 locally (+ confirmation about cent7 earlier)

PR #91, opened 2020-04-30
Add support for Ubuntu 20.04

corny:


lae:
I am absolutely lost right now.

Inserted a test into CI for this PR, but it's causing several other distributions to fail: https://travis-ci.org/github/lae/ansible-role-netbox/builds/683767355

The deploys work fine, but the resulting NetBox deployment returns HTTP 500 errors (some issue with pickle). Thinking that maybe master itself had broke, I ran another build without these changes, but it passed just fine: https://travis-ci.org/github/lae/ansible-role-netbox/builds/683867367. And on the off chance it had to do with disk space, I removed Xenial as a target in this build but it still failed: https://travis-ci.org/github/lae/ansible-role-netbox/builds/683821015

I really can't see how these changes would cause this sort of cascading failure right now so I might have to postpone merging this until later.

stappersg:
FWIW   consider that _travis ci_  lacks _ubuntu 20.04_ support

Yeah, that be could be total nonsense. It is only base on my first encounter with _travis ci_ where I had to **down-grade** `auto-tools` to get a succesfull build ...

lae:
NetBox isn't being tested on Ubuntu 20.04 via Travis CI, it's being tested in a Ubuntu 20.04 LXC container (on a Ubuntu 18 Travis CI image).

lae:
Merging as-is since I don't see anything wrong with the changeset and it passes CI if Ubuntu Focal is the only distro being tested: https://travis-ci.org/github/lae/ansible-role-netbox/builds/740362417

Will be foregoing routine testing for Focal for now however since adding Focal to CI causes all other distro tests to fail (per earlier comment).

Issue #103, opened 2020-07-02
Problem with molecule test


Shm013:

Shm013:
lae.netbox❯ molecule --version
molecule 3.0.4
   ansible==2.9.10 python==3.7

Shm013:
It might be because of old version of molecule
https://molecule.readthedocs.io/en/latest/configuration.html#lint

PR #106, opened 2020-07-25
Clean up ansible-lint and yamllint issues

Shm013:
Issue: [ #103](https://github.com/lae/ansible-role-netbox/issues/103)

Lint OK, but `ansible test` still failed.


lae:
Can you squash your commits into one and overwrite this branch (force push) with that commit?

Shm013:
@lae Done

.travis.yml
```diff
@@ -1,3 +1,5 @@
+---
```
lae, line 1: Can you ignore this file for the linter? It's not exactly part of Ansible (and I don't think ansible-lint checks it?), and the Travis tool formats it without indents so it'll likely get massively reformatted again on edits.

examples/netbox_config.yml
```diff
@@ -17,22 +17,22 @@ netbox_config:
       - Dale Gribble
       - dgribble@example.com
   # note that the array is nested above
-  BANNER_BOTTOM: *BANNER_TOP
+  BANNER_BOTTOM: '*BANNER_TOP'
```
lae, line 20: Does this change the value? It's a reference to another value.

Shm013: Is this thing passes to jinja2? If so it must be the same.

lae: What do you mean?

My understanding is that Ansible parses the YAML first, so here I think it was supposed to reference the value of the BANNER_TOP key and that value itself would get passed to Jinja for templating.

Shm013: Hmm... I don't know how it working. Where does the variable for `BANNER_BOTTOM` come from?

There is a problem if leave everything as is (without quotes).

```
❯ ansible-lint
Warning: Failed to load examples/netbox_config.yml with We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  found undefined alias

The error appears to be in '/home/shm/Documents/projects/ansible-role-netbox/examples/netbox_config.yml': line 20, column 18, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  # note that the array is nested above
  BANNER_BOTTOM: *BANNER_TOP
                 ^ here
, assuming is not a playbook.
```

lae: Looks like the reference creation needs to be placed further above. I think? e.g.:

```
  BANNER_TOP: &BANNER_TOP 'Your banner text'
  BANNER_BOTTOM: *BANNER_TOP
```

(but also, it's an example so I'm not sure if the linter should be parsing it)

Issue #66, opened 2019-12-12
SECRET_KEY - autogeneration broken


cherusk:
From reading the source, intended behaviour is, if SECRET_KEY is not in netbox_config parameter, it will be generated. But 

```
      - include_role:
          name: "netbox"
        vars:
          netbox_stable: true
          netbox_database_host: localhost
          netbox_database_user: "netbox"
          netbox_database_password: "netbox"
          netbox_superuser_password: "netbox"
          netbox_socket: "0.0.0.0:9000"
          netbox_load_initial_data: false
          netbox_application_log: "file:{{ netbox_shared_path }}/application.log"
          netbox_requests_log: "file:{{ netbox_shared_path }}/requests.log"
          netbox_scripts: []
          netbox_config:
             ALLOWED_HOSTS:
               - 0.0.0.0
             MEDIA_ROOT: "{{ netbox_shared_path }}/media"
             REPORTS_ROOT: "{{ netbox_shared_path }}/reports"
             SCRIPTS_ROOT: "{{ netbox_shared_path }}/scripts"
          #   SECRET_KEY: "" # auto-generated
```
Is ailing with:
```
    qemu: TASK [_netbox : Run database migrations for NetBox] ****************************
[...]

PR #108, opened 2020-09-01
Use 'private' variable for netbox_config (#66)

beddari:
This works around the problem in #66 that if you use the (increasingly more
common) include_role pattern variables you set externally could appear
immutable to the role logic. Thus, using set_fact with a combine() to
add SECRET_KEY to netbox_config will not work as expected.

Introducing a new 'private' (and thus mutable) variable in
load_variables works around the problem and allows the include_role
pattern to work.


lae (approved):

lae:
https://travis-ci.org/github/lae/ansible-role-netbox/builds/723059611

PR #111, opened 2020-10-15
Add support for database OPTIONS dictionary

### LINKED ISSUE #104: Add support for database OPTIONS dictionary
Similar to the discussion in https://github.com/lae/ansible-role-netbox/issues/64 (and realizing I put my own foot in my mouth because here we are) there are some use cases where it's up to the user's environment. In my particular case, I need to set the `sslmode` parameter to something other than the default.

I think the simplest option would be to introduce a new var like:

`netbox_database_options` that would be a dictionary used similar to `netbox_uwsgi_options` and `netbox_config` are already handled today.

For cases like [isolation levels](https://docs.djangoproject.com/en/2.2/ref/databases/#isolation-level) we could add a note that the _real_ numerical value must be used instead of the constant `psycopg2.extensions.ISOLATION_LEVEL_SERIALIZABLE` as specified in the example. It seems like the other [possible options](https://www.postgresql.org/docs/10/libpq-connect.html#LIBPQ-PARAMKEYWORDS) for PostgreSQL are either strings or numbers so the existing logic used for `netbox_config` should work fine here as well.

A quick pass at the template logic

```jinja
DATABASE = {
    'NAME': '{{ netbox_database }}',
    'USER': '{{ netbox_database_user }}',
{% if netbox_database_host is defined %}
    'PASSWORD': '{{ netbox_database_password }}',
    'HOST':  '{{ netbox_database_host }}',
    'PORT': '{{ netbox_database_port }}',
{% else %}
    {% if netbox_database_password is defined %}
    'PASSWORD': '{{ netbox_database_password }}',
    {% endif %}
    'HOST': '{{ netbox_database_socket }}',
{% endif %}
    'CONN_MAX_AGE': {{ netbox_database_conn_age }},
{% if netbox_database_options is defined %}\
    'OPTIONS': {
        {% for setting, value in netbox_database_options.items() %}
        {% if value is string or value is number %}
        "{{ setting }}": {{ value | to_nice_json }},
        {% else %}
        "{{ setting }}": json.loads(r'''{{ value | to_json }}'''),
        {% endif %}
        {% endfor %}
    }
{% endif %}
}
```

tyler-8:
Fixes #104

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -163,6 +163,16 @@ set `netbox_database_conn_age` to your preferred maximum connection age, in seco
 
 [source,yaml]
 ----
+# Example usage, default is empty dict
+netbox_database_options:
+  sslmode: require
+  isolation_level: 3
+----
+
+If you need to set any other PostgreSQL parameter key words you can do so here. For cases like https://docs.djangoproject.com/en/3.1/ref/databases/#isolation-level[isolation levels] the numerical value must be used instead of the constant: `psycopg2.extensions.ISOLATION_LEVEL_SERIALIZABLE` vs `3`. Only add things here if you really know what you're doing.
+
+[source,yaml]
+----
 netbox_redis_host: 127.0.0.1
 netbox_redis_port: 6379
 netbox_redis_password: ''
--- a/README.adoc
+++ b/README.adoc
@@ -169,7 +169,11 @@ netbox_database_options:
   isolation_level: 3
 ----
 
-If you need to set any other PostgreSQL parameter key words you can do so here. For cases like https://docs.djangoproject.com/en/3.1/ref/databases/#isolation-level[isolation levels] the numerical value must be used instead of the constant: `psycopg2.extensions.ISOLATION_LEVEL_SERIALIZABLE` vs `3`. Only add things here if you really know what you're doing.
+If you need to set any other PostgreSQL parameter key words you can do
+so here. For cases like https://docs.djangoproject.com/en/3.1/ref/databases/#isolation-level[isolation levels] the numerical value must be used
+instead of the constant:
+`psycopg2.extensions.ISOLATION_LEVEL_SERIALIZABLE` vs `3`.
+Only add things here if you really know what you're doing.
 
 [source,yaml]
 ----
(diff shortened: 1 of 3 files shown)


lae:
Is this still a WIP?

tyler-8:
No it isn't, just forgot to take it out of draft mode, woops!

README.adoc
```diff
@@ -161,6 +161,16 @@ To configure Netbox to keep database connections open longer than a single reque
...
 
+[source,yaml]
+----
+# Example usage, default is empty dict
+netbox_database_options:
+  sslmode: require
+  isolation_level: 3
+----
+
+If you need to set any other PostgreSQL parameter key words you can do so here. For cases like https://docs.djangoproject.com/en/3.1/ref/databases/#isolation-level[isolation levels] the numerical value must be used instead of the constant: `psycopg2.extensions.ISOLATION_LEVEL_SERIALIZABLE` vs `3`. Only add things here if you really know what you're doing.
```
lae, line 172: This needs to be word-wrapped (79 characters is what I use b/c of PEP-8).

templates/configuration.py.j2
```diff
@@ -28,6 +28,15 @@ DATABASE = {
     'HOST': '{{ netbox_database_socket }}',
 {% endif %}
     'CONN_MAX_AGE': {{ netbox_database_conn_age }},
+{% if netbox_database_options is defined and netbox_database_options|length > 0 %}
```
lae, line 31: Checking for definition here isn't really necessary since it's always defined via defaults, but I guess it doesn't hurt.

lae (commented):
Looks fine overall. This might be a good chance to try adding the `#jinja2: lstrip_blocks: True` you mentioned earlier, though, and see how the resulting configuration turns out.

Also, populating a group variable in `tests/group_vars/netbox` for this with an option or two would be a good idea.

README.adoc
tyler-8: Corrected. Though the line with the link is longer, but is obviously shorter after rendering.

templates/configuration.py.j2
tyler-8: Removed.

tyler-8:
I thought I'd already resolved this PR long ago, sorry about that. Should be good to go now.

lae (approved):

PR #119, opened 2020-12-14
Netbox 2.10 needs manage.py trace_paths run on upgrade

mtinberg:
Looking at the upgrade.sh for Netbox 2.10 it needs to run manage.py trace_paths --no-input once to populate the new dcim_cablepath table, the manage.py migrate step doesn't do it.  In addition there are a couple of other commands in the update script to invalidate caches that probably make sense to run on updates.

### DIFF
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -122,6 +122,12 @@
         app_path: "{{ netbox_current_path }}/netbox"
         virtualenv: "{{ netbox_virtualenv_path }}"
 
+    - name: Trace missing cable paths for NetBox
+      django_manage:
+        command: "trace_paths --no-input"
+        app_path: "{{ netbox_current_path }}/netbox"
+        virtualenv: "{{ netbox_virtualenv_path }}"
+
     - name: Create a super user for NetBox
       shell: "printf '{{ netbox_superuser_script }}' |\
               {{ netbox_virtualenv_path }}/bin/python {{ netbox_current_path }}/netbox/manage.py shell"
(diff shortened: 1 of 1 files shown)


tyler-8 (approved):
LGTM.

These are good tasks to better match the official upgrade script, and the path trace is necessary for v2.10

tasks/deploy_netbox.yml
```diff
@@ -122,6 +122,13 @@
         app_path: "{{ netbox_current_path }}/netbox"
         virtualenv: "{{ netbox_virtualenv_path }}"
 
+    - name: Trace missing cable paths for NetBox
+      django_manage:
+        command: "trace_paths --no-input"
+        app_path: "{{ netbox_current_path }}/netbox"
+        virtualenv: "{{ netbox_virtualenv_path }}"
+      when: "netbox_stable_version is version('2.10.0', '>=')"
```
lae, line 130: Can you also add a check for the git commit hash that introduces this change? You can reference commit 192137b9834ee21e979b6ec4d32da7cc6bd729ea to see how it's been implemented previously.

tasks/deploy_netbox.yml
```diff
@@ -142,5 +149,18 @@
...
+      django_manage:
+        command: "remove_stale_contenttypes --no-input"
+        app_path: "{{ netbox_current_path }}/netbox"
+        virtualenv: "{{ netbox_virtualenv_path }}"
+
+    - name: Clear cached data in NetBox
+      django_manage:
+        command: "invalidate all"
+        app_path: "{{ netbox_current_path }}/netbox"
+        virtualenv: "{{ netbox_virtualenv_path }}"
```
lae, line 162: Will these two commands work for older versions of NetBox/Django (even if they might not do anything)? If they work for like the releases in the past year or two then they're fine, otherwise it might be appropriate to also have a version check here.

lae (approved):
Overall lgtm. I'll merge in a few days if I don't hear back (and check the things I mentioned in the review).

mtinberg:
sorry for the late response, just cleaning up email at end of week

I can git blame the upgrade.sh script

As far as the trace_paths looks like it was introduced in 2.10.0 https://github.com/netbox-community/netbox/commit/f560693748d1f35e79ad9d006a8a9b75ef5ae37b

Looks like the cache invalidation commands were added for Netbox 2.6.13 https://github.com/netbox-community/netbox/commit/685cf50268858c3ea211907ba63a69638c9cef8f

—
Mark Tinberg ***@***.***>
Division of Information Technology-Network Services
University of Wisconsin-Madison


________________________________________

tyler-8:
@mtinberg - if you were to add those version/hash checks that @lae mentioned on those two tasks, this PR should be good to go.

tasks/deploy_netbox.yml
lae: f560693748d1f35e79ad9d006a8a9b75ef5ae37b

tasks/deploy_netbox.yml
lae: Appears to be fine.

Issue #132, opened 2021-07-07
Role needs to be updated to latest NetBox


lae:
Hey guys. I haven't managed to proactively keep this role in sync with upstream (living circumstances over the past year and a half haven't been very conducive to many activities), so I could use some help if anyone's willing to pitch in.

Basically, the major version release notes and all minor version update notes for NetBox need to be reviewed and then this role needs to be updated to support new features or use updated configurations, etc. Some changes that might affect earlier versions negatively may need to be version-gated, as well. 

This should be done separately for the latest 2.10 and latest 2.11 releases. Refer to the following PRs for examples of how it's been done in the past.

- https://github.com/lae/ansible-role-netbox/pull/27
- https://github.com/lae/ansible-role-netbox/pull/40
- https://github.com/lae/ansible-role-netbox/pull/61
- https://github.com/lae/ansible-role-netbox/pull/67
- https://github.com/lae/ansible-role-netbox/pull/87
- https://github.com/lae/ansible-role-netbox/pull/115

Please feel free to open a PR like this to keep track of your work if you'd like to work on this. I'd greatly appreciate it. Whenever possible, I'll try to provide my own code reviews as well.

In the worst case that this isn't able to be completed sooner, I'll probably be able to focus more on this hopefully by the end of the year.

Dzordzu:
I can confirm that role (as of now) works perfectly for 2.10 and 2.11 (I'm currently using netbox 2.11.11)
But now 3.0.1 is present, so it may break

lae:
I'm not *too* surprised that it works as-is for later versions, but if there were new features introduced they need configuration support within the role.

TheMonkeyBrain:
Just did an update from Netbox v2.11.11 to v3.1.9 by using the changes in PR #134 and #136 and didn't run into any issues.

tyler-8:
I did an upgrade from 2.10.10 to 3.1.11 using #141 and it went smoothly.

MozeBaltyk:
Guy may I suggest to add to the readme as prequisite to make this role work ? (for those who use ansible-core) 

```bash
ansible-galaxy collection install community.postgresql   
```

Issue #127, opened 2021-03-04
Netbox 2.11 planning to drop support for python36


mtinberg:
https://github.com/netbox-community/netbox/issues/5532

I checked on my CentOS 7 host and rh-python38 exists as a software collection so we should be able to change the package install portion to require that and use it to create the venv `scl enable rh-python38 "python -m venv py38venv"` which may fix some of the other issue with newer cryptography and rust.

I haven't tested on Ubuntu 18.04 and this change is a few months out, hopeful in the interim we can rework the playbook to get the newer python runtime installed on all the platforms where it is readily available.

lae:
Are software collections typically installable out of the box on CentOS? Last I recall they weren't/involved extra administration work to setup installing from software collections, but my memory is hazy.

If not, I'm not sure it would be appropriate to have this role take on the scope of doing so. It might just be best to write documentation/example playbook for deploying to CentOS using another role or something, and have this role test for/expect Python 3.8.

mtinberg:
That's very reasonable, we should just have a constraint check for python38 and have a message with some clue what to do when it's not found

—
Mark Tinberg ***@***.***>
Division of Information Technology-Network Services
University of Wisconsin-Madison


________________________________________

lae:
Fixed in #141

Issue #125, opened 2021-02-18
cryptography NAPALM dependancy needs rust and newer pip on python36


mtinberg:
Netbox doesn't deploy on RHEL/CentOS 7 python36 with pip3 v9.0.3 due to current release of cryptography depended on by NAPALM depending on rust during the build/install process, pip3 v9.0.3 setuptools doesn't have modules to build rust, but running `pip3 install pip --upgrade` in the venv then installing napalm works, as long as rust (1.47.0) is installed from EPEL.

We may need to add rust as a package dependancy then upgrade pip in the venv during deployment, maybe have the first pip/venv task be to set up an empty venv and upgrade pip, then install requirements/constraints then optional modules.

This may affect other platforms as well

tyler-8:
This has affected a lot of packages/apps in the Python ecosystem.

I'm not crazy about blindly upgrading pip to the latest version - as there have been breaking `pip` changes in the past. IMO, we should select a known working version (whatever is released right now) and pin that as part of the pip constraints perhaps?

The latest version of pip has a pre-compiled wheel of the cryptography package so there's no need to install `rust` on the target system. So the pre-install steps should be `pip install pip==X.Y.Z wheel=X.Y.Z` to ensure the following tasks can complete.

tyler-8:
It's worth noting that the NetBox `upgrade.sh` is doing a 'blind' upgrade now https://github.com/netbox-community/netbox/issues/5776 but I still think a pinned `pip` is the better approach.

lae:
I'm on board with pinned `pip` via the constraints role variable. Although pinning to a minor version (x.y) instead of a patch version (x.y.z) seems like it'd be appropriate to keep the maintenance overhead low.

Issue #124, opened 2021-02-10
deploy_netbox.yml is missing trace_paths


michaljbarczewski:
Netbox 2.10 added new handling of cables, including cable paths caching,
this requires generating paths for existing cables during upgarde otherwise cable tracing does not work for existing cables.

Default upgrade.sh script runs the following command:
COMMAND="python3 netbox/manage.py trace_paths --no-input"

This should probably be added to deploy_netbox.yml

michaljbarczewski:
I now see that there is already a pull request (#119 ) which address this issue, so this can be closed.

PR #134, opened 2021-10-09
Update netbox version to 3.0.7

rsp2k:


lae:
Hi. Thanks for your PR.

Have you gotten a chance to review https://github.com/lae/ansible-role-netbox/issues/132 and if so, can you confirm that you've read through the changelogs for NetBox up until the current release to ensure that any new configuration changes/features are accounted for in this changeset?

lae:
So I quickly tried to validate for myself and read through the changes from 2.9.9 to 2.10.10 (so there might still be stuff missing for 2.11.x and 3.0.x), and found these two that affect this role:
https://github.com/netbox-community/netbox/issues/5938 (which we already have an issue for, #127, and I guess #125 is also of concern)
https://github.com/netbox-community/netbox/issues/6083 (see associated PR for details)

I also realised I could quickly do a diff on some files to help pinpoint configuration changes and upgrade.sh changes from 2.9.9 to current:

https://gist.github.com/lae/a24a0bc366e76626ff65f48c9ed32319

On a quick skim most of the config file changes look superficial and so the only thing that could probably use an update is the example at `examples/netbox_config.yml`. The redis change (the second PR mentioned above) is done in the config template/with role variables. One of the upgrade changes already has a PR at #124 that I guess I need to go back to later.

PR #138, opened 2021-11-19
Python 3.8 support for Netbox 3.x on RHEL 7

mtinberg:
I have reworked the playbook to use Python 3.8 from SCL on RHEL7 but haven't tested on other platforms such as Debian/Ubuntu or RHEL8. 

There are some workarounds for not using a system-wide install of Python with Ansible and uWSGI, Ansible builtin template module imports the selinux library which does not seem to have a prebuilt package for SCL Python3.8 on RHEL7 the way that libselinux-python and libselinux-python3 support python27 and python36 on RHEL7, so I reset the ansible_python_interpreter around sections that use ansible.builtin.template when the OS is RHEL7 (or similar like CentOS or OracleLinux).  I also allowed it to install the selinux library in the virtualenv but to use that as the ansible_python_interpreter would require building the venv earlier and further restructuring. 

On that vein I did explicitly build the venv in its own command rather than having it implicitly happen as part of the first pip install as the implicit venv doesn't seem to upgrade and existing python36 venv to python38, it uses the version of python that the venv was originally built with.  It may be more clear to just remove the existing venv before the first pip command, if you try to upgrade python using the same version of netbox, unlike how it'll build a new venv when building a new version of netbox. 

The other workaround was to install uWSGI in the Python3.8 virtualenv and reference that in the systemd service file instead of trying to install it systemwide in /usr/local/bin/uwsgi as the systemwide one doesn't upgrade from python36 to python38 or have the right path information built into it when using Python3.8 from SCL on RHEL7.  So on my system at least I end up with three versions of uwsgi, one installed from RPM in /usr/sbin/uwsgi, one installed from the playbook in python36 in /usr/local/bin/uwsgi and one installed from the playbook with python38 in /opt/netbox/current/venv-py3/bin/uwsgi  

Take a look at the diff and tell me what you think, is this something you want?

### DIFF
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -190,6 +190,13 @@
         - netbox_superuser_create_token
         - netbox_superuser_enabled
 
+    - name: Build the local documentation
+      command:
+        chdir: "{{ netbox_current_path }}"
+        cmd: "{{ netbox_virtualenv_path }}/bin/mkdocs build"
+      when:
+        - netbox_stable and netbox_stable_version is version('3.0', '>=')
+
     - name: Generate static assets for NetBox
       django_manage:
         command: "collectstatic --no-input"
@@ -216,12 +223,6 @@
         app_path: "{{ netbox_current_path }}/netbox"
         virtualenv: "{{ netbox_virtualenv_path }}"
 
-    - name: Build the local documentation
-      command:
-        chdir: "{{ netbox_current_path }}"
-        cmd: "{{ netbox_virtualenv_path }}/bin/mkdocs build"
-      when:
-        - netbox_stable and netbox_stable_version is version('3.0', '>=')
 
   become: true
   become_user: "{{ netbox_user }}"
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -161,6 +161,7 @@
         app_path: "{{ netbox_current_path }}/netbox"
         virtualenv: "{{ netbox_virtualenv_path }}"
       retries: 2
+      throttle: 1
       register: _netbox_db_migration
       until: _netbox_db_migration is success
 
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -85,9 +85,10 @@
     owner: "{{ netbox_user }}"
     group: "{{ netbox_group }}"
     mode: 0640
-    validate: "{{ netbox_virtualenv_path }}/bin/python -c \"import py_compile,os; f=r'%s';\
-               c='/tmp/' + os.path.basename(os.path.dirname(f)) + '-' + os.path.basename(f) + 'c';\
-               py_compile.compile(f, c); os.remove(c)\""
+    validate:
+      "{{ netbox_virtualenv_path }}/bin/python -c \"import py_compile,os; f=r'%s';\
+      c='/tmp/' + os.path.basename(os.path.dirname(f)) + '-' + os.path.basename(f) + 'c';\
+      py_compile.compile(f, c); os.remove(c)\""
   notify:
     - reload netbox.service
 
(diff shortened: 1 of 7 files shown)


lae:
If I don't get a chance to look at this by the end of the month, feel free to ping me a reminder.

Of course, if anyone else is available, please review this (and possibly test on other distros) and give me your thoughts.

br-olf:
@lae I'll take the duty to ping you regarding this issue.
I currently fight with RHEL8 and would love to see a merge.

I'll checkout the merge request for myself and give you an update soonish but I'll be on vacation for 2 weeks so updates from myself will have to wait until may.

mtinberg:
No worries, this has continued to work for me although I did make one change to run the mkdocs command _before_ collecting the static files as it doesn't work the other way around 😉



—

Mark Tinberg ***@***.***>

Division of Information Technology-Network Services

University of Wisconsin-Madison


________________________________

ironiq:
@mtinberg just a note: i cannot see anywhere installing the "centos-release-scl" package on CentOS 7 to be able to install rh-python38 packages. I think the "centos-sclo-*" repos are not enabled by default everywhere. Or am i missing something?

mtinberg:
I added a task to install_packages_yum.yml which I'll commit as soon as my test in the lab completes

```yaml
- name: Install SCL repository (CentOS)
  yunm:
    name: centos-release-scl
    state: installed
    enablerepo: extras
  when: "ansible_distribution == 'CentOS' and ansible_distribution_major_version == '7'"
```


—

Mark Tinberg ***@***.***>

Division of Information Technology-Network Services

University of Wisconsin-Madison


________________________________

tyler-8:
IMO - any tasks related to installation of a Python binary need to be optional. Not every CentOS user wants to or can install the EPEL or `centos-release-scl`. Adding something like the existing var `netbox_install_epel`, such as `netbox_install_centos_scl: False`.

ironiq:
@tyler-8 how will you provide python3.8 other way on redhat7-based distros? IMO building the package locally is even worse.

tyler-8:
Ultimately, it doesn't matter if the binary comes from EPEL, SCL, a private repo, a local build-from-source, or any other means, it should be outside the scope of this role. As long as I provide a valid Python binary that meets the compatibility requirements of NetBox, this role should function.

All I'm suggesting in this instance is to make repository installs optional.

mtinberg:
I get it, wrap the SCL install (like EPEL) in some optional flags, so the admin can install their own python37+ and that choice won't be overridden by installing epel/scl/etc.  So we'll need to make sure we don't presume the path of the installed python as well, if the admin is providing it outside of this playbook


—

Mark Tinberg ***@***.***>

Division of Information Technology-Network Services

University of Wisconsin-Madison


________________________________

tyler-8:
> So we'll need to make sure we don't presume the path of the installed python as well, if the admin is providing it outside of this playbook

I believe you should be fine as long as you're using `netbox_python_binary` and not hard-coding any paths (outside of defaults)

mtinberg:
I've configured this path in redhat-7.yml as
_netbox_python_binary: /opt/rh/rh-python38/root/usr/bin/python3.8
 and I'd have to re-read the whole thing to make sure that isn't a bad way to do it as I don't recall offhand how data flows through this particular playbook and if there is a better place to configure this (make it conditional on whether a netbox_enable_scl: true flag is thrown elsewhere in the config like for epel).

—

Mark Tinberg ***@***.***>

Division of Information Technology-Network Services

University of Wisconsin-Madison


________________________________

mtinberg:
There is bit of weirdness if I recall right because when ansible logs into the system it may be using the /usr/bin/python or /usr/bin/python3 interpreter but neither of those can manipulate the venv for python37+ using native modules without changing which interpreter ansible itself is using on the managed system, so there is a lot of switching back and forth the ansible_python variable, esp when the system-wide install comes with the selinux python module that ansible expects to use when manipulating files, but the scl rh-python38 doesn't come with that for some reason.

anyway I'd need to make sure that an admin-supplied python is used in the same way and it doesn't accidently switch to an scl installed python that doesn't exist when it's not enabled.


—

Mark Tinberg ***@***.***>

Division of Information Technology-Network Services

University of Wisconsin-Madison


________________________________

PR #142, opened 2022-03-25
generate offline documentation during deployment

karolyczovek:


tasks/deploy_netbox.yml
```diff
@@ -134,6 +134,14 @@
         - netbox_stable and netbox_stable_version is version('2.10.0', '>=')
           or netbox_git and _netbox_git_contains_trace_paths.rc == 0
 
+    - name: Build the local documentation
+      shell:
+        cmd: "{{ netbox_virtualenv_path }}/bin/mkdocs build"
+        chdir: "{{ netbox_current_path }}"
+      when:
+        - netbox_stable and netbox_stable_version is version('3.1.0', '>=')
+          or netbox_git and _netbox_git_contains_trace_paths.rc == 0
```
lae, line 143: `_netbox_git_contains_trace_paths.rc` refers to a specific git commit where the `trace_paths` command was introduced.

Please see commit 8addd275253d50792db98309b146896d4ea717bd for how this works. We basically identify the commit X feature was introduced, and version gate it via that commit. Can you update this to refer to the relevant commit in upstream (it should probably be a merge commit from the relevant PR)?

lae (changes requested):

tasks/deploy_netbox.yml
karolyczovek: Hmm.. even cable trace is not needed, will update PR in a minute

karolyczovek: should be fine now, generate docs if >3.1.0 or git

lae: People using `netbox_git` are not guaranteed to be on top of Netbox' development branch (they could be using their own fork). It still needs to be version gated via the relevant commit.

karolyczovek: ok, will try to lookup which comit introduced the local doc generation in netbox upstream

karolyczovek: 7058d6ca5ae2901383874bcea5fedad31e165dca will be our friend

karolyczovek: changes implemeted

lae (approved):
LGTM, but I want to double check on one thing. The last commit changes the version gate from 3.1.0 to 2.12.0. It looks like the mkdocs commit was added to 3.0.0, so shouldn't it be set to that? (and 2.12.0 doesn't look like it exists) What's the reasoning behind that?

karolyczovek:
> LGTM, but I want to double check on one thing. The last commit changes the version gate from 3.1.0 to 2.12.0. > It looks like the mkdocs commit was added to 3.0.0, so shouldn't it be set to that? (and 2.12.0 doesn't look like >it exists) What's the reasoning behind that?

Might be i interpret wrong, https://github.com/netbox-community/netbox/issues/6328 mentions docs generation as 2.11 feature - this was the reasoning behind.

lae:
Nah, the version specified in that issue is just what the current version of Netbox was when that person opened a feature request. If you look at the commit, https://github.com/netbox-community/netbox/commit/7058d6ca5ae2901383874bcea5fedad31e165dca, it'll show you what tags it is under and the earliest seems to be 3.0-beta1. Anyway, merging this, thanks.

PR #143, opened 2022-04-05
Ammend requirments

cfleeelcocks:
added ansible replace to requirements.txt.when it contains links to specific repository branches.


lae (changes requested):
Hello, and thanks for your PR!

Firstly, there are two dozen commits that appear to be extraneous and specific to your environment. Please rebase your branch against the main working branch and force push it to update this PR.

Secondly, can you provide an example of the issue this is supposed to resolve? You need to provide an instance of this occurring in upstream Netbox.

PR #141, opened 2022-03-23
Netbox 3.0 release

smutel:
Fix #132 

Add tasks to support installation of Netbox 3.0:
* Use latest pip/setuptools in virtualenv => create first the virtualenv with latest pip/setuptools before installing other dependencies
* Change permissions on netbox home => when installing pip libraries the `{{ netbox_home }}/.cache` will be used so `{{ netbox_home }}` need to be owned by netbox user
* Update for Netbox **3.0** => some tasks has been marked only for netbox <= 3.0 / include https://github.com/lae/ansible-role-netbox/pull/139
* Setup housekeeping (https://github.com/lae/ansible-role-netbox/pull/134)

Some ideas/information were retrieved from:
* https://github.com/lae/ansible-role-netbox/pull/136
* https://github.com/lae/ansible-role-netbox/issues/132

PR #137, opened 2021-10-31
NetBox 2.10 release

lae:
This introduces feature support for NetBox 2.10.

## Changes

* `INSECURE_SKIP_TLS_VERIFY` under redis config ([2.10.9 changelog](https://github.com/netbox-community/netbox/releases/tag/v2.10.9))
* trace_paths management command ([2.10.0 changelog](https://github.com/netbox-community/netbox/releases/tag/v2.10.0))
* Vagrant playbook updates since CI isn't working for now

PR #148, opened 2022-09-03
Run on v3.3.2

oscarcarlsson:


lae:
Major version bumps in this role need to support any new features/configuration changes in upstream. Please see #132. More recent examples include #137 and #141.

oscarcarlsson:
Ah, I will update this PR when I have time!

PR #155, opened 2023-01-12
Allow optional install of uwsgi into Netbox venv, and v3.2 version constraint

### LINKED ISSUE #144: Install uwsgi in venv and point service file to venv binary
Currently, the `uwsgi` installation isn't done into a virtual environment, but the environment of `ansible_python_interpreter`, which may not be using the same python binary that the NetBox venv is using.

This task installs `uwsgi` and happens _after_ the NetBox-specific venv has been created.

https://github.com/lae/ansible-role-netbox/blob/dbead6819c0bd65aecc511404453e6b9c210ab7a/tasks/main.yml#L68-L79

I propose adding `virtualenv: "{{ netbox_virtualenv_path }}"` to that task to ensure it's installed using the same venv as the rest of the app.

Additionally, the systemd service template relies on `/usr/bin/env`, which doesn't have any awareness of the venv as currently written.

https://github.com/lae/ansible-role-netbox/blob/dbead6819c0bd65aecc511404453e6b9c210ab7a/templates/netbox.service.j2#L8-L9

If the above task is modified to install `uwsgi` into the venv, then the template could be adjusted to use

```
ExecStart={{ netbox_virtualenv_path }}/bin/uwsgi --ini {{ netbox_shared_path }}/uwsgi.ini`
```

All in all this would give users more flexibility in using Python versions from different sources (altinstall, pyenv, package managers, etc).

Something would probably need to be modified with this task as well but we'd likely need to decouple it from `netbox_python_binary` (and possibly other tasks).

https://github.com/lae/ansible-role-netbox/blob/dbead6819c0bd65aecc511404453e6b9c210ab7a/tasks/install_packages_yum.yml#L28-L35

tyler-8:
Fixes #144 

Also added a new version constraint for NetBox v3.2 requiring Python 3.8

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -419,6 +419,15 @@ Specify extra configuration options to insert into `uwsgi.ini` here. This is
 expected to be a dictionary of key/value pairs, e.g. `buffer-size: 65535`.
 
 [source,yaml]
+netbox_uwsgi_in_venv: false
+
+Toggle `netbox_uwsgi_in_venv` to `true` if you want `uwsgi` to be installed in the same virtual environment as NetBox.
+Otherwise, it will be installed system-wide into the library path of the python version used to created the virtual environment (normal/legacy behavior).
+
+WARNING: There's a possibility that this may become the default in a later version of this role (I think after further cross-platform testing).
+See https://github.com/lae/ansible-role-netbox/issues/144[issue #144] for further details.
+
+[source,yaml]
 netbox_install_epel: true
 
 Toggle `netbox_install_epel` to `false` if you do not want this role to install
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -70,6 +70,7 @@
     name: uwsgi
     state: "{{ 'latest' if netbox_keep_uwsgi_updated else 'present' }}"
     umask: "0022"
+    virtualenv: "{{ omit if not netbox_uwsgi_in_venv else netbox_virtualenv_path }}"
   environment:
     PATH: "/usr/local/bin:{{ _path }}"
   notify:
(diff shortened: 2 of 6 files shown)


tasks/main.yml
```diff
@@ -77,6 +77,22 @@
   retries: 2
   register: _netbox_uwsgi_install
   until: _netbox_uwsgi_install is succeeded
+  when: "not netbox_uwsgi_in_venv | bool"
+
+- name: Install uWSGI via pip (inside venv, new behavior)
```
lae, line 82: I'm not sure this task needs to be duplicated. Unless  `omit` doesn't work in the `virtualenv` parameter?

i.e. `virtualenv: "{{ netbox_uwsgi_in_venv | ternary(netbox_virtualenv_path, omit) }}"` (going off on memory, double check this syntax)

README.adoc
```diff
@@ -418,6 +418,13 @@ netbox_uwsgi_options: {}
 Specify extra configuration options to insert into `uwsgi.ini` here. This is
 expected to be a dictionary of key/value pairs, e.g. `buffer-size: 65535`.
 
+[source,yaml]
+netbox_uwsgi_in_venv: false
+
+Toggle `netbox_uwsgi_in_venv` to `true` if you want `uwsgi` to be installed in
+the same virtual environment as NetBox. Otherwise it will install using the Python
+version used to created the virtual environment (normal/legacy behavior).
```
lae, line 426: update phrasing here since it's a bit confusing to me

"...will be installed system-wide, into the library path of the python version used to..."

lae (changes requested):

tasks/main.yml
p-rintz: I have tested the following 
`    virtualenv: "{{ omit if not netbox_uwsgi_in_venv else netbox_virtualenv_path }}"`
to work.

lae (approved):

PR #158, opened 2023-01-29
Add `netbox_database_maintenance` variable

MonsterDruide1:
Some use cases might not be able to access the default `postgres` database. To make sure that this playbook also works in these scenarios, `maintenance_db` should be changed by using variables, passed from outside of the role.

The default should still remain on `postgres`:
https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_db_module.html

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -169,6 +169,13 @@ set `netbox_database_conn_age` to your preferred maximum connection age, in seco
 
 [source,yaml]
 ----
+netbox_database_maintenance: postgres
+----
+
+If the postgres database is configured to only allow access to specific tables of the DB for the user configured with Netbox, you can set `netbox_database_maintenance` to replace the default database used for connection checking to a different table than the default `postgres`. This is an empty table in every postgres database by default, but some configurations might block access to this table, so a different table (i.e. `netbox_prod`) can be used here instead.
+
+[source,yaml]
+----
 # Example usage, default is empty dict
 netbox_database_options:
   sslmode: require
@@ -501,6 +508,7 @@ installing NetBox on to authenticate with it over TCP:
     netbox_database: netbox_prod
     netbox_database_user: netbox_prod_user
     netbox_database_password: "very_secure_password_for_prod"
+    netbox_database_maintenance: netbox_prod
     redis_bind: 127.0.0.1
     redis_version: 6.0.9
     redis_checksum: sha256:dc2bdcf81c620e9f09cfd12e85d3bc631c897b2db7a55218fd8a65eaa37f86dd
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -46,6 +46,7 @@
     name: "{{ netbox_database }}"
     login_user: "{{ netbox_database_user }}"
     login_unix_socket: "{{ netbox_database_socket }}"
+    maintenance_db: "{{ netbox_database_maintenance }}"
   become: true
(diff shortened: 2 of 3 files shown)


tasks/main.yml
```diff
@@ -59,6 +59,7 @@
     port: "{{ netbox_database_port }}"
     login_user: "{{ netbox_database_user }}"
     login_password: "{{ netbox_database_password }}"
+    maintenance_db: "{{ netbox_database_maintenance }}"
```
lae, line 62: Is this the only place that needs to reference it?

MonsterDruide1: With this setup, it works for us locally... it might be useful to add to the socket connection method above as well, give me a minute to push that change as well.

lae: `templates/configuration.py.j2` also references database connection information, but if it's working for you without an edit there then I guess it's probably fine....

MonsterDruide1: Done. Can't test it though, so ... careful with the change higher in the file.

lae: If you have Vagrant, you can actually test using vagrant. Otherwise I guess I'll try to test it, but I think it's probably fine in the first place if it's a default.

lae (approved):

lae (changes requested):
Actually, almost forgot. Please update the `README.adoc` to include documentation for the new role variable and its usage.

MonsterDruide1:
I have added it ... somewhere, not quite sure where/how you'd want me to document it.

lae:
There's a section that goes over the other database connection parameters. That would be where. Ctrl+F "netbox_database" should get you to it, I think, if you're not sure where.

lae:
@MonsterDruide1 I think you might have forgot to force push the documentation update, so pinging you just in case. (Also see my last comment if you need it.)

MonsterDruide1:
Sorry, took some time with other stuff - does this documentation match your request?

README.adoc
```diff
@@ -167,6 +167,13 @@ To configure Netbox to keep database connections open longer than a single reque
 set `netbox_database_conn_age` to your preferred maximum connection age, in seconds.
 300 seconds (5 minutes) is typically a good number to start with.
 
+[source,yaml]
+----
+netbox_database_maintenance: netbox_prod
```
lae, line 172: Sorry, this should show what the default is set to (so `postgres` instead of `netbox_prod`)

README.adoc
```diff
@@ -501,6 +508,7 @@ installing NetBox on to authenticate with it over TCP:
     netbox_database: netbox_prod
     netbox_database_user: netbox_prod_user
     netbox_database_password: "very_secure_password_for_prod"
+	netbox_database_maintenance: netbox_prod
```
lae, line 511: should be 4 spaces instead of a tab

lae:
Yeah, the documentation looks fine apart from the minor things I pointed out.

lae (approved):

README.adoc
MonsterDruide1: Alright, changed this one, but left the other one - that makes sense in the given example, as it uses a "known-to-be-existing-and-accessible" database table.

lae: Yeah, the other one is actually using it in an example playbook to override it, so it's fine. It's just the Role Variables section where defaults are shown (to save having to look at the actual defaults files).

Issue #129, opened 2021-05-17
Support for Deployment of Multiple RQ Workers


tyler-8:
It's possible, particularly now with plugins, to have a need for multiple `rq-workers` in the environment to process all the background tasks simultaneously. The default one-worker deployment could fall behind if multiple reports, scripts, and other tasks are queued at the same time. Plugins can introduce even more complex workflows that may take longer to run. 

The good news is `rq` supports deploying multiple workers quite easily and without making any code changes to NetBox itself. The `systemd` method of doing so is documented in the [official docs](https://python-rq.org/patterns/systemd/).

> to start workers by invoking `systemctl start rqworker@1.service`, systemctl start rqworker@2.service from the terminal.
> 
> You can also reload all the workers by invoking `systemctl reload rqworker@*`.

I'm imagining a simple setting like `netbox_rqworker_count = 1` as the default. Users could change it to `4`, which would start & enable `rqworker@1`, `rqworker@2`, `rqworker@3`, `rqworker@4`.

PR #161, opened 2023-07-24
Allow use of multiple rqworkers

p-rintz:
This closes issue #129 
I oriented myself based on the [netbox-rq.service](https://github.com/netbox-community/netbox/blob/develop/contrib/netbox-rq.service) of the main netbox project.
I removed the reload of the rqworker and used a restart instead, as the service file from netbox does not have a reload.

Default install stayed at 1 process.

- Allow use of multiple rqworkers
- Add tests for changed rqworker processes

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -239,7 +239,7 @@ webhooks database.
 netbox_rqworker_processes: 1
 ----
 
-Specify how many rqworkers should be started by the systemd service.
+Specify how many request queue workers should be started by the systemd service.
 You can leave this at the default of 1, unless you have a large number of reports,
 scripts and other background tasks.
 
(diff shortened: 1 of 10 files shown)


README.adoc
```diff
@@ -234,6 +234,15 @@ Use this syntax if your redis is installed with sentinet architecture (multiple
 the second set of variables if you wish to split your cache database from your
 webhooks database.
 
+[source,yaml]
+----
+netbox_rqworker_processes: 1
+----
+
+Specify how many rqworkers should be started by the systemd service.
```
lae, line 242: might be better to unpack "rqworkers" into "request queue workers" for readers here.

lae (approved):
Content of the changes looks okay to me. I can't test it out atm, but perhaps @tyler-8 might want to take a look at this?

README.adoc
p-rintz: Sounds good. Changed.

PR #162, opened 2023-09-22
Compatibility with Debian 12

davidmehren:
The newer Python on Debian 12 requires the use of the `--break-system-packages` pip flag to install packages globally.

Debian 12 also does not ship with a cron implementation anymore, so the `systemd-cron` compatibility package needs to be installed.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -72,6 +72,7 @@
     name: uwsgi
     state: "{{ 'latest' if netbox_keep_uwsgi_updated else 'present' }}"
     umask: "0022"
+    extra_args: "{{ (ansible_distribution_release == 'bookworm') | ternary('--break-system-packages', '') }}"
   environment:
     PATH: "/usr/local/bin:{{ _path }}"
   notify:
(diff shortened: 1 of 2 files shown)


lae:
Long term I think we want to not pip install uwsgi into /usr/lib.
Instead of `--break-system-packages`, per latest merge commit can you test `netbox_uwsgi_in_venv: true` works fine under Debian 12? We could just force installation into the virtualenv for Debian 12 as starters for moving to a virtualenv-only deployment.

sleinen:
I have tried @lae's suggestion in [comment 1750519004](https://github.com/lae/ansible-role-netbox/pull/162#issuecomment-1750519004):

Instead of adding `--break-system-packages` (which does sound scary), I added

```
netbox_uwsgi_in_venv: true
```

to @davidmehren's `debian-12.yml`, and with that I was able to install NetBox just fine on Debian, based on a recently forked NetBox git repo (3.6.5+).  (I did have to perform a small piece of Git configuration, for which I'll create a separate issue.)

lae:
Superseded by #163.

PR #163, opened 2023-11-05
Compatibility with Debian 12, venv-based

sleinen:
Based on @davidmehren's [PR #162](https://github.com/lae/ansible-role-netbox/pull/162), but implementing @lae's [suggestion](https://github.com/lae/ansible-role-netbox/pull/162#issuecomment-1750519004) to install uWSGI in a venv instead of using `--break-system-packages`.


lae (approved):

lae:
Thanks @sleinen!

PR #165, opened 2023-11-05
Run "git log" under the same user that cloned the repo

### LINKED ISSUE #164: Git-based installation: problematic file ownership
When using the `netbox_git: True` installation option, the Ansible role will clone the source repo (`netbox_git_uri`) as user `netbox_user`.  It then goes on to look for a number of commits in the Git history to set various variables used later in the code, e.g. `_netbox_git_contains_issue_2239_fix.rc`, using (roughly) `git log | $commit_id`.  This is currently done as root.

On newer Git, when `git log` is called as `root` in a working tree owned by non-root (e.g. user `netbox`), it will complain:

```
fatal: detected dubious ownership in repository at '/srv/netbox/releases/git-repo'
To add an exception for this directory, call:

	git config --global --add safe.directory /srv/netbox/releases/git-repo
```

These error messages are normally hidden from Ansible's output, but the effect is that all checks return a non-zero exit code (due to the `pipefail`), which causes issues further downstream, such as configuration mistakes based on erroneous assumptions about the state of the NetBox code.

Declaring `{{ netbox_git_repo_path }}` as a safe directory in user `root`'s Git configuration would be one way to fix this issue.  Alternatively, we could run the `git log` commands as the same router that owns the working tree.

sleinen:
This avoids "fatal: detected dubious ownership in repository" errors that cause all feature-detection commands to fail.

Fixes #164

### DIFF
--- a/tasks/install_via_git.yml
+++ b/tasks/install_via_git.yml
@@ -15,77 +15,82 @@
     group: "{{ netbox_group }}"
     state: directory
 
-- name: Check existence of commit 1fb67b7, fixing issue netbox#2239
-  shell: 'set -o pipefail; git log --format=%H "{{ netbox_git_version }}" | grep ^1fb67b791f1a91c624dae4a1cd256e4cf3ddbb77'
-  args:
-    chdir: "{{ netbox_git_repo_path }}"
-    executable: /bin/bash
-  register: _netbox_git_contains_issue_2239_fix
-  changed_when: false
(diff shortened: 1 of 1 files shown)


lae (changes requested):
Instead of adding the `become` parameters to each task, can you move the check tasks into their own block and use the `become` parameters on the block itself? (This does mean indenting all of the check tasks.)

lae (approved):

PR #166, opened 2024-01-10
Compatibility with Netbox 3.7

davidmehren:
https://github.com/netbox-community/netbox/commit/2d1f88272497ca72d2e1eca8e291c04538c6810e removed the 'clearcache' command, so this needs to be skipped for release 3.7.0 and later.


lae (approved):
lgtm

sleinen:
Thanks @davidmehren! This works fine in the "stable" case, but not quite when installing from Git.  See #168

PR #182, opened 2024-06-03
fix: Adding handling of dictionaries and variable merging

melonger:
- Using a method to nest Dictionaries, that don't require duplication of defaults.  Merging the default values recursively into the answer file if custom values no defined.

- Added daemon reload for better immutability when "re-deploying" in a tainted environment.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -50,16 +50,7 @@ netbox_metrics_enabled: false
 netbox_metrics_dir: netbox_metrics
 netbox_metrics_path: "/run/{{ netbox_metrics_dir }}"
 
-netbox_config:
-  # SECRET_KEY:
-  ALLOWED_HOSTS:
-    - localhost
-    - 127.0.0.1
-  # NAPALM_USERNAME:
-  # NAPALM_PASSWORD:
-  MEDIA_ROOT: "{{ netbox_shared_path }}/media"
-  REPORTS_ROOT: "{{ netbox_shared_path }}/reports"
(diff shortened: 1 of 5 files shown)


defaults/main.yml
```diff
@@ -50,7 +50,8 @@ netbox_metrics_enabled: false
 netbox_metrics_dir: netbox_metrics
 netbox_metrics_path: "/run/{{ netbox_metrics_dir }}"
 
-netbox_config:
+netbox_config: {}
+netbox_config_defaults:
```
lae, line 54: Can you move this to a "private" variable (`_netbox_default_config`) in `vars/main.yml`? (this is sort of the convention we use)

tasks/validate_variables.yml
```diff
@@ -42,13 +42,14 @@
...
   when: "netbox_napalm_enabled | bool"
 
-- name: NetBox user content directories must be specified
-  assert:
-    that:
-      - "'MEDIA_ROOT' in netbox_config"
-      - "'REPORTS_ROOT' in netbox_config"
-      - "'SCRIPTS_ROOT' in netbox_config"
-    msg: "Please ensure MEDIA_ROOT/REPORTS_ROOT/SCRIPTS_ROOT are defined in netbox_config."
+# Defaulted using Dictionary inheritance via load_variables:32
```
lae, line 45: Please just remove this task entirely. It'll still be in git's history so no need to comment this out. (line number being referenced is always subject to change, too.)

lae (changes requested):

lae (approved):
addressed previous review with rebase

Issue #157, opened 2023-01-15
Implement Custom Validators


tyler-8:
I propose adding support for `Custom Validators` using the [dotted path](https://docs.netbox.dev/en/stable/customization/custom-validation/#dotted-path) configuration method so users can still take advantage of the Dynamic Configuration support.

I'm already working on a branch to develop this but wanted to get an issue in place.

lae:
Seems like this was introduced in 3.1.0? (https://github.com/netbox-community/netbox/issues/7619) So from my perspective this is one of those features that should be configurable by this role in order to get 3.1.0 support through the door in the first place.

If I understand correctly this would fundamentally be the same as `netbox_scripts`/`netbox_reports` (with the slight difference of those validators being referenced in the custom validators config)?

tyler-8:
They were part of the [3.0 release](https://docs.netbox.dev/en/stable/release-notes/version-3.0/#custom-model-validation-5963). They are slightly different from reports and scripts because there is no "root" folder that NetBox automatically looks for them in.

That means the user has to have a package/folder in the path so NetBox can followed the dotted path configured (I use `local_validators` in the same directory as `manage.py`).

Other than that, I am using similar logic to reports/scripts for copying & linking files.

lae:
Oh I thought the custom validator support in 3.0 was just a specific DSL in `configuration.py`, not external files.

I just thought of something I've done before that might be useful. You can add directories to the PYTHONPATH at runtime:

``` 
import os
import sys
sys.path.insert(0, os.path.join(os.getcwd(), "{{ shared_directory }}/validators"))
```

That could go at the top of the `configuration.py` template. This way they don't necessarily have to be in the netbox folder itself, and might possibly be simpler to reason about (e.g. what if someone wanted to use a different package name than `local_validators` in their config?). Food for thought.

tyler-8:
> specific DSL in `configuration.py`, not external files.

It's both :) 

The external files part is where you want to write a CustomValidator class that overrides the `validate()` method to do more complex validation than what the DSL provides. The [dynamic configuration](https://docs.netbox.dev/en/stable/configuration/#dynamic-configuration-parameters) is "best" IMO and so there's no need to put anything in `configuration.py` (except perhaps your `sys.path` trick to get the path set up correctly) - and if users of this role want/need to define their validators in `configuration.py`, they still can using the role as it exists today, rather than adding new role-specific variables to handle it.

sol1-matt:
[...]

PR #190, opened 2024-11-26
Support for local_settings.py settings override file

sol1-matt:
This PR configures local_settings.py.

This is required for the deployment of the branch plugin as this file must exist before the migrations are run if the plugin is enabled. You can't otherwise use this role if you want the branch plugin.

- The role now automatically tries to copy from `{{ playbook_dir }}/files/netbox/local_settings.py`, files/netbox as local_settings.py is a generic name that could be used by other plays or roles.
- Errors are ignored on the copy in place of testing if the files exists or a var being set
- The symlink creation tests for the existence of local_settings.py in the shared director. If a user add's local_settings.py to the shared directory manually the role will still use it.
- The role doesn't have a way to remove the local_settings.py from the shared directory, if the user wants to remove the local_settings.py they can remove it from shared.

### DIFF
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -133,6 +133,30 @@
   notify:
     - reload netbox.service
 
+# local_settings.py
+- name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
+  copy:
+    src: "{{ playbook_dir }}/files/netbox/local_settings.py"
+    dest: "{{ netbox_shared_path }}/local_settings.py"
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+  ignore_errors: yes
+
+- name: Check if local_settings.py file exists in shared
+  stat:
+    path: "{{ netbox_shared_path }}/local_settings.py"
+  register: netbox_local_settings_file_in_shared
+
+- name: Symlink/Remove NetBox local_settings.py file into/from the active NetBox release
+  file:
+    src: "{{ netbox_shared_path + '/local_settings.py' if netbox_local_settings_file_in_shared.stat.exists else omit }}"
+    dest: "{{ netbox_config_path }}/local_settings.py"
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+    state: "{{ 'link' if netbox_local_settings_file_in_shared.stat.exists else 'absent' }}"
+  notify:
+    - reload netbox.service
+
 - name: Copy NetBox scripts into SCRIPTS_ROOT
   copy:
     src: "{{ item.src }}"
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -136,23 +136,23 @@
 # local_settings.py
 - name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
   ansible.builtin.copy:
-    src: "{{ netbox_local_settings_config_file }}"
+    src: "{{ netbox_local_settings_file }}"
     dest: "{{ netbox_shared_path }}/local_settings.py"
     owner: "{{ netbox_user }}"
     group: "{{ netbox_group }}"
   ignore_errors: yes
   when:
-    - netbox_local_settings_enabled
+    - netbox_local_settings_file is defined
   notify:
     - reload netbox.service
 
 - name: Symlink/Remove NetBox local_settings.py file into/from the active NetBox release
   file:
-    src: "{{ netbox_shared_path + '/local_settings.py' if netbox_local_settings_enabled else omit }}"
+    src: "{{ netbox_shared_path + '/local_settings.py' if netbox_local_settings_file is defined else omit }}"
     dest: "{{ netbox_config_path }}/local_settings.py"
     owner: "{{ netbox_user }}"
     group: "{{ netbox_group }}"
-    state: "{{ 'link' if netbox_local_settings_enabled else 'absent' }}"
+    state: "{{ 'link' if netbox_local_settings_file is defined else 'absent' }}"
   notify:
     - reload netbox.service
 
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -135,25 +135,24 @@
 
 # local_settings.py
 - name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
-  copy:
-    src: "{{ playbook_dir }}/files/netbox/local_settings.py"
+  ansible.builtin.copy:
+    src: "{{ netbox_local_settings_config_file }}"
     dest: "{{ netbox_shared_path }}/local_settings.py"
     owner: "{{ netbox_user }}"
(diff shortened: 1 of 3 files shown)


tasks/deploy_netbox.yml
```diff
@@ -133,6 +133,30 @@
   notify:
     - reload netbox.service
 
+# local_settings.py
+- name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
+  copy:
+    src: "{{ playbook_dir }}/files/netbox/local_settings.py"
```
lae: can you change this to be a role variable, e.g. `netbox_local_settings_file` (if you have a better name feel free to suggest it), and use the variable as the basis for whether or not this file exists and gets copied or not? e.g.

```yaml
# minimal example, I'm not suggesting to remove any attributes here
- name: copy local_settings
  ansible.builtin.copy:
    src: "{{ netbox_local_settings_file }}"
    dest: "{{ netbox_shared_path }}/local_settings.py"
  when:
    - netbox_local_settings_file is defined
```

that way you don't need the additional `stat` task, nor do you need to use `ignore_errors`, and can use `{{ 'link' if netbox_local_settings_file is defined else 'absent' }}` instead in the symlink task.

we also don't want to just assume people know this feature exists, so having it in the role variables adds at the very least some visibility and configurability in case they prefer a different directory layout. that said, the role variable should also be added to the README with an explanation of its purpose afterwards.

tasks/deploy_netbox.yml
```diff
@@ -133,6 +133,30 @@
   notify:
     - reload netbox.service
 
+# local_settings.py
+- name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
+  copy:
```
lae: Please use fully qualified module names for any new/modified tasks. (`ansible.builtin.copy`)

tasks/deploy_netbox.yml
```diff
@@ -133,6 +133,30 @@
...
+  ignore_errors: yes
+
+- name: Check if local_settings.py file exists in shared
+  stat:
+    path: "{{ netbox_shared_path }}/local_settings.py"
+  register: netbox_local_settings_file_in_shared
+
+- name: Symlink/Remove NetBox local_settings.py file into/from the active NetBox release
+  file:
+    src: "{{ netbox_shared_path + '/local_settings.py' if netbox_local_settings_file_in_shared.stat.exists else omit }}"
```
lae: This can be simplified afterwards too.

tasks/deploy_netbox.yml
```diff
@@ -133,6 +133,30 @@
...
     - reload netbox.service
 
+# local_settings.py
+- name: Copy NetBox local_settings.py into shared (ignore if it doesn't exist)
+  copy:
+    src: "{{ playbook_dir }}/files/netbox/local_settings.py"
+    dest: "{{ netbox_shared_path }}/local_settings.py"
+    owner: "{{ netbox_user }}"
+    group: "{{ netbox_group }}"
+  ignore_errors: yes
```
lae: this task should also have a notify handler, not just the symlink task, because if the `local_settings.py` file changes on a later run the symlink doesn't get recreated and the notify handler from it won't execute either.

lae (changes requested):
Please see comments.

for own reference,
background on what the `local_settings.py` file is and why one would use it: https://github.com/netbox-community/netbox/issues/16127

tasks/deploy_netbox.yml
sol1-matt: The reason for the `ignore_errors` was if you have more than one netbox, say production and dev, and the local_settings.py needs to be different for each you can skip the role deployment of the different files and do that step as a `pre_task`.

The 2 ways I can see to get this aligned with what your suggesting and not lock it into a single file for all Netbox instances in a repo is
* keep `icinga_errors` and `stat` task and add `netbox_local_settings_enabled` boolean so you can use a `pre_task` to deploy the file to shared and the role will handle the symlinking
* add a `netbox_local_settings_file` string which is the filename that needs to be picked up and saved to `local_settings.py` then we don't need `icinga_errors: false` or the `stat` task but if you don't put the file in the right place it all fails (which seems fair enough)
```
netbox_local_settings_file: "local_settings_prod.py"
```

I think the second option is probably more sensible thinking about it more

sol1-matt: the second pattern would make more sense for #157 too, having consistency for end users on "how to deploy custom files that aren't templates" is nice

lae: the second option is exactly what I'm requesting. an admin should voluntarily enable the variable, and configure it differently per environment based on their own playbook layout/deployment pipelines (but we can suggest a sensible value for them to use, which is the one you already have in the PR).

sol1-matt: looking at ldap I should bring this in line with the vars for that except it is file not template
```
netbox_ldap_enabled: false
netbox_ldap_config_template: netbox_ldap_config.py.j2
```

lae: Hmm...
In the case of LDAP, there are other tasks specific to supporting LDAP asides from just copying the template over, such as injecting library dependencies into the list of venv packages, etc, so it simplifies things that part of the process.
I think for just a single file copy it'd be annoying to have two variables for it, unless there's additional configuration that needs to be done? The `is defined` test works well enough for a single task imo.

Definitely makes sense to use a boolean toggle for #157 though, since I guess you will need to modify `configuration.py.j2` to use it.

sol1-matt:
I've [committed changes](https://github.com/sol1-matt/ansible-role-netbox/commit/dcc474966e4a3097bfecf5ff9c2a2c3593a13c41) but still need to test them.

tasks/deploy_netbox.yml
sol1-matt: I'll make this one a single var then and set the default to empty

sol1-matt:
@lae I've tested this now and updated the readme, the last commit for this is 282202a894e251eea9608a168b76d415a6afedf6. 
It is ready for merge.

lae (approved):

lae:
Just for future reference `is defined` is used to test if a variable is defined. Since we set an empty default it'll still error like below:

```
TASK [lae.netbox : Copy NetBox local_settings.py into shared (ignore if it doesn't exist)] ***************************************************************************************************
fatal: [10.17.0.122]: FAILED! => {"changed": false, "msg": "src (or content) is required"}
...ignoring

TASK [lae.netbox : Symlink/Remove NetBox local_settings.py file into/from the active NetBox release] *****************************************************************************************
fatal: [10.17.0.122]: FAILED! => {"changed": false, "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /srv/netbox/shared/local_settings.py", "path": "/srv/netbox/current/netbox/netbox/local_settings.py", "src": "/srv/netbox/shared/local_settings.py"}
```

We have a couple other role variables that only get used if they're defined, but just left commented out in the documentation and defaults file, so I applied that pattern here.

lae:
I just noticed: congrats on being the 100th PR on this project!

PR #191, opened 2024-11-29
2024.11.29 miscellaneous fixes

lae:
- **[docs] make wrapping more in-line with asciidoc's semantic convention**
- **move config file deployment tasks into a single block + lint**
- **Set ansible-galaxy namespace/role_name**
- **omit src attribute in symlink tasks during removal**


lae:
who would ever want to review my PRs? 😔

Issue #205, opened 2025-10-28
Optionally skip housekeeping cron setup task


tyler-8:
[NetBox v4.4 replaced the housekeeping script](https://github.com/netbox-community/netbox/issues/18349) with a NetBox-managed/scheduled jobs.

This task will need another set of conditionals to check if it should be run. 

https://github.com/lae/ansible-role-netbox/blob/0bcea2e0b3eda35d8e6c720fb188dd1fff599bb3/tasks/deploy_netbox.yml#L195

PrakharJain1509:
@tyler-8 Can I work on this issue?

tyler-8:
> [@tyler-8](https://github.com/tyler-8) Can I work on this issue?

@PrakharJain1509 go for it :)

PR #206, opened 2025-10-30
Optionally skip housekeeping cron; auto-remove for NetBox v4.4+

PrakharJain1509:
### PR Description
- NetBox v4.4 replaces the legacy housekeeping script with NetBox-managed scheduled jobs. This PR ensures we don’t create a stale cron entry on v4.4+ and allows operators to disable the cron explicitly.

- Changes:
  - Add `netbox_housekeeping_cron_enabled` (default: `true`) to control whether the legacy housekeeping cron is managed.
  - Only create the cron for stable installs when `netbox_stable_version >= 3.0.0` and `< 4.4.0`; for git installs, rely on the existing commit presence check.
  - Add cleanup task to remove the cron when not applicable (disabled or v4.4+), preventing leftovers during upgrades.

- Files edited:
  - `tasks/deploy_netbox.yml`: Add version-aware guard, toggle, and cleanup task for the housekeeping cron.
  - `defaults/main.yml`: Introduce `netbox_housekeeping_cron_enabled` variable.

- Backwards compatibility:
  - No breaking changes. Existing behavior is preserved for NetBox < 4.4 unless the new toggle is set to `false`.

- Upgrade notes:
  - Upgrading to NetBox v4.4+ will automatically remove the legacy housekeeping cron entry.
  - Operators can explicitly disable cron creation by setting:
    - `netbox_housekeeping_cron_enabled: false`

- Testing:
  - Lint clean.
  - Verified behavior across scenarios:
    - Stable < 4.4: cron is created when enabled.
    - Stable >= 4.4: cron is not created and is removed if present.
    - Git: cron managed when the housekeeping commit exists and toggle is enabled.
    - Toggle off: cron is absent across all versions.

- Example override:
  - `group_vars`:
    - `netbox_housekeeping_cron_enabled: false`
 For PR #205

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -97,3 +97,8 @@ netbox_keep_uwsgi_updated: false
 netbox_uwsgi_in_venv: false
 netbox_uwsgi_options: {}
 netbox_rqworker_processes: 1
+
+# Whether to install the legacy housekeeping cron job.
+# NetBox v4.4+ replaces the housekeeping script with NetBox-managed scheduled jobs.
+# Set to false to skip creating the cron entry regardless of version.
+netbox_housekeeping_cron_enabled: true
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -97,8 +97,3 @@ netbox_keep_uwsgi_updated: false
 netbox_uwsgi_in_venv: false
 netbox_uwsgi_options: {}
 netbox_rqworker_processes: 1
-
-# Whether to install the legacy housekeeping cron job.
-# NetBox v4.4+ replaces the housekeeping script with NetBox-managed scheduled jobs.
-# Set to false to skip creating the cron entry regardless of version.
-netbox_housekeeping_cron_enabled: true
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -200,8 +200,25 @@
     user: "{{ netbox_user }}"
     cron_file: "netbox"
   when:
-    - netbox_stable and netbox_stable_version is version('3.0.0', '>=')
-      or netbox_git and _netbox_git_contains_housekeeping.rc == 0
+    - netbox_housekeeping_cron_enabled | bool
+    - (netbox_stable and netbox_stable_version is version('3.0.0', '>=')
+        and netbox_stable_version is version('4.4.0', '<'))
+      or (netbox_git and _netbox_git_contains_housekeeping.rc == 0)
+
+- name: Remove housekeeping cronjob when not applicable
+  cron:
+    name: "Netbox housekeeping"
+    user: "{{ netbox_user }}"
+    cron_file: "netbox"
+    state: absent
+  when:
+    - not (
+        (netbox_housekeeping_cron_enabled | bool)
+        and (
+          (netbox_stable and netbox_stable_version is version('3.0.0', '>=') and netbox_stable_version is version('4.4.0', '<'))
+          or (netbox_git and _netbox_git_contains_housekeeping.rc == 0)
+        )
+      )
 
 - block:
     - name: Run database migrations for NetBox
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -192,33 +192,20 @@
     group: "{{ netbox_group }}"
   loop: "{{ netbox_reports }}"
 
-- name: Schedule daily housekeeping cronjob
+- name: Manage daily housekeeping cronjob
   cron:
     name: "Netbox housekeeping"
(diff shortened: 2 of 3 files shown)


PrakharJain1509:
@tyler-8 Please have a look into this and let me know if you need anything else

lae (changes requested):
Looking over this, a separate task should not be used.

Also, the same treatment should apply to git deployments that contain the upstream change. Specifically, https://github.com/netbox-community/netbox/pull/19815. See `install_via_git.yml` for examples.

In fact, if my understanding after looking through that PR is correct, we don't exactly need a role variable to managing the cronjob? It should just exist or not exist depending on what NetBox version is installed, so the version comparison would be enough...though probably in the `state` attribute instead.

tyler-8:
> Looking over this, a separate task should not be used.
> 
> Also, the same treatment should apply to git deployments that contain the upstream change. Specifically, [netbox-community/netbox#19815](https://github.com/netbox-community/netbox/pull/19815). See `install_via_git.yml` for examples.
> 
> In fact, if my understanding after looking through that PR is correct, we don't exactly need a role variable to managing the cronjob? It should just exist or not exist depending on what NetBox version is installed, so the version comparison would be enough...though probably in the `state` attribute instead.

It looks like the new task is only for users currently using this playbook and upgrading to v4.4 - it cleans up the previously installed cron file. Otherwise, as I've tested myself, the housekeeping cron file will persist unless you manually delete it, so I think that's a reasonable addition.

However, I do agree that a new configurable variable isn't necessary. We should do a version comparison or look for the presence of the relevant commit ID and register the variable to determine if action is needed just like already happens here: https://github.com/lae/ansible-role-netbox/blob/0bcea2e0b3eda35d8e6c720fb188dd1fff599bb3/tasks/install_via_git.yml#L72

defaults/main.yml
```diff
@@ -97,3 +97,8 @@ netbox_keep_uwsgi_updated: false
 netbox_uwsgi_in_venv: false
 netbox_uwsgi_options: {}
 netbox_rqworker_processes: 1
+
+# Whether to install the legacy housekeeping cron job.
+# NetBox v4.4+ replaces the housekeeping script with NetBox-managed scheduled jobs.
+# Set to false to skip creating the cron entry regardless of version.
+netbox_housekeeping_cron_enabled: true
```
tyler-8: We don't need a new variable.

tyler-8 (changes requested):
Instead of introducing `netbox_housekeeping_cron_enabled`, we should instead extend the already existing task that sets `_netbox_git_contains_housekeeping` to include checking for the commit ID that now removes the houskeeping feature.

However, the cleanup task is a good idea I think.

lae:
> It looks like the new task is only for users currently using this playbook and upgrading to v4.4 - it cleans up the previously installed cron file.

Right, but I'm suggesting that cleanup doesn't need a new task. Setting the `state` attribute on the existing task to a conditional should be enough to clean it up. Unless I'm severely misunderstanding something...

tyler-8:
> > It looks like the new task is only for users currently using this playbook and upgrading to v4.4 - it cleans up the previously installed cron file.
> 
> Right, but I'm suggesting that cleanup doesn't need a new task. Setting the `state` attribute on the existing task to a conditional should be enough to clean it up. Unless I'm severely misunderstanding something...

I see what you mean now and agreed - dynamically adjusting the `state` attribute should be sufficient in the existing task.

lae:
@PrakharJain1509 can we get an update on this?

lae:
https://github.com/lae/ansible-role-netbox/blob/2e3435e24d815294d0c6817ce269ee1a31bc4a3f/tasks/install_via_git.yml#L75

https://github.com/netbox-community/netbox/commit/f3a8b9c1d2e4f5a6b7c8d9e0f1a2b3c4d5e6f7a8b9c0

\>commit doesn't exist

:thinking:

PrakharJain1509:
@lae  @tyler-8 Did some changes..

lae:
I don't know if GitHub intends to fix this, but I will just mention that my previous comment/review is a response to Jain and is the most recent comment and the timestamp for Jain's was moved an hour ahead for some reason. (note the commit and comment/gha notification timestamps are basically an hour apart, give a minute for writing the comment)

image

which means it's time for our annual mention of https://www.youtube.com/watch?v=-5wpm-gesOY

[image](https://www.youtube.com/watch?v=-5wpm-gesOY)

lae:
Closing due to lack of update and as far as I understand something hallucinated an illegitimate SHA. Please reopen if you honestly intend to see this contribution through to completion.

PR #208, opened 2026-04-01
feat: setuptools version pin

tsingletonacic:
Allow users to pin setuptools version, useful for maintaining older installs

### DIFF
--- a/tasks/deploy_netbox.yml
+++ b/tasks/deploy_netbox.yml
@@ -24,8 +24,8 @@
   ansible.builtin.pip:
     name:
       - pip
-      - setuptools
-    state: latest
+      - "setuptools{{ netbox_setuptools_version }}"
+    state: present
     virtualenv: "{{ netbox_virtualenv_path }}"
     virtualenv_command: "{{ netbox_python_binary }} -m venv"
   become: true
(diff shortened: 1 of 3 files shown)


tasks/deploy_netbox.yml
```diff
@@ -24,8 +24,8 @@
   ansible.builtin.pip:
     name:
       - pip
-      - setuptools
-    state: latest
+      - "setuptools{{ netbox_setuptools_version|default('') }}"
```
lae: the default filter here is redundant, I believe?

tsingletonacic: It is a second safety just in case, but it can be removed indeed, I will let you decide.

Thanks

lae: Okay, as far as I understand it is dead code and should be removed.

lae: @tsingletonacic did you actually want me to make the change itself? I would prefer it if you amended and force pushed the commit yourself so that the line stays credited to you.

Anyway I'll come back to this in like a week or so in case you're busy.

tsingletonacic: Sorry, misunderstood your comment. I have fixed in my fork.

lae (approved):

PR #209, opened 2026-05-13
Add support for new 'DATABASES' configuration keyword since Netbox 4.3

eliba:
This supersedes the 'DATABASE' keyword, which is still supported but marked as deprecated. Noticed this when I tried to install a plugin on a relatively recent version of Netbox (4.5.2).

I'm not really happy with indentation in `configuration.py` yet.
* Syntax < 4.3:
   ```
   DATABASE = {
       'NAME': 'netbox',
       'USER': 'netbox',
       'HOST': '/var/run/postgresql',
       'CONN_MAX_AGE': 0,
   }
   ```
* Syntax >= 4.3:
   ```
   DATABASES = {
   'default': {
       'ENGINE': 'django.db.backends.postgresql',
       'NAME': 'netbox',
       'USER': 'netbox',
       'HOST': '/var/run/postgresql',
       'CONN_MAX_AGE': 0,
       }
   }
   ```
The official documentation shows a nicer indentation, but it's hard to template without duplicating the whole `DATABASE` block in the template, suggestions welcome: https://netboxlabs.com/docs/netbox/installation/netbox/#databases

### DIFF
--- a/templates/configuration.d/databases.j2
+++ b/templates/configuration.d/databases.j2
@@ -1,0 +1,25 @@
+    'NAME': '{{ netbox_database }}',
+    'USER': '{{ netbox_database_user }}',
+{% if netbox_database_host is defined %}
+    'PASSWORD': '{{ netbox_database_password }}',
+    'HOST':  '{{ netbox_database_host }}',
+    'PORT': '{{ netbox_database_port }}',
+{% else %}
+{%      if netbox_database_password is defined %}
+    'PASSWORD': '{{ netbox_database_password }}',
+{%      endif %}
+{%      if netbox_database_alt_socket and netbox_database_socket is defined %}
+    'PORT': '{{ netbox_database_port }}',
+{%      endif %}
+    'HOST': '{{ netbox_database_socket }}',
+{% endif %}
+    'CONN_MAX_AGE': {{ netbox_database_conn_age }},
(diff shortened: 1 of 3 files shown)


lae (approved):
This looks functional and fine for now if we need to get this out sooner than later.

However, maybe it's possible to dynamically set an indentation value based on these conditions, and then use the `indent` Jinja2 filter against all strings? Alternatively, maybe we could use a child template for the inner configuration, and indent the templated result dynamically. I'll let you decide if you want to try tackling it one of these ways or not.

templates/configuration.d/databases.j2
```diff
@@ -0,0 +1,34 @@
+{% if netbox_stable and netbox_stable_version is version('4.3', '>=') or
+      netbox_git and _netbox_git_contains_database_rename.rc == 0 %}
+'default': {
```
lae: If you move this "outer block" to the calling template, I think there should really be no need to test this set of conditionals multiple times? i.e.

```
{% test %}
'default': {
    'ENGINE': 'pg',
{%   filter %}
{%   include %}
{%   endfilter %}
}
{% else %}
{% include %}
{% endif %}
```

eliba: Yes good point, I adapted it this way.

lae (approved):

